PR #3, opened 2017-04-25
Publishing and changelog updates

brentertz:
Add "release:semantic" and "changelog" npm scripts and developer publishing docs


kylegach:
👍

docs/publishing.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Semantic
+
+When publishing a public release, you should be able to simply run the following command.  It will do many things, like make a production build, run tests, detect which packages have changed, determine appropriate version bumps, generate change-logs, add git tags, push to github, create a release on github, and publish to npm.
+
+* `npm run release:semantic`
+
+## Manual
+
+If you need a bit more control over the process, use the following command as your starting point.
```
limscoder: Will we ever need this or should we always be using `release:semantic` ?

brentertz: We should usually use `release:semantic`, but there are cases such as when doing prereleases that we will need the latter.  The docs are informal at this point and are so that it is not in solely my head.

PR #15, opened 2017-05-08
docs - add component READMEs

coldpour:
### Description
each component README has 
* a link to the top level README
* installation steps with npm
* sample jsx showing the usage of the component

### Motivation and context
the current README files were empty and need something in them so that people can consume our components

### How has this been tested?
i created a react app and pulled in all the components and tested that they showed up in the browser.

### Types of changes

### Checklist

i wish i had a way to template these READMEs...

### DIFF
--- a/packages/hello-world/README.md
+++ b/packages/hello-world/README.md
@@ -1,1 +1,26 @@
 # HelloWorld
+
+Mineral UI `HelloWorld` component. Find out more about [Mineral UI](https://github.com/mineral-ui/mineral-ui).
+
+## Installation
+
+```sh
+npm install --save @mineral-ui/hello @mineral-ui/world @mineral-ui/hello-world
+```
+
+## Usage
+
+```jsx
+import React, { Component } from 'react';
+import HelloWorld from '@mineral-ui/hello-world';
+
+export default class MyComponent extends Component {
+  render() {
+    return (
+      <div>
+        <HelloWorld />
+      </div>
+    );
+  }
+}
+```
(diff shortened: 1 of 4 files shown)


packages/hello-world/README.md
```diff
@@ -1 +1,26 @@
...
+
+```sh
+npm install --save @mineral-ui/hello @mineral-ui/world @mineral-ui/hello-world
+```
+
+## Usage
+
+```jsx
+import React, { Component } from 'react';
+import HelloWorld from '@mineral-ui/hello-world';
```
brentertz: This component actually requires that you import both the hello and world components separately as they are peer dependencies.

packages/hello-world/README.md
```diff
@@ -1 +1,26 @@
...
+
+export default class MyComponent extends Component {
+  render() {
+    return (
+      <div>
```
brentertz: Do we want the extra wrapping div on these examples?

brentertz:
We want to keep the Archetype readme as it will serve as the template for starting new components in the future.

packages/hello-world/README.md
coldpour: it requires that you npm install them all but you don't actually have to import all of them

packages/hello-world/README.md
coldpour: I think so since it makes it a more realistic example. I doubt people will be making their own components that simply render ours. I guess they could want to create a component that just changes the prop signature... but that seems far fetched to me.

packages/hello-world/README.md
brentertz: Derp, you are right.

Issue #45, opened 2017-06-06
[Theme] Variable/naming schema


Mikewaite:
As a Mineral Developer, we need to create a system of variables that tackles size, colors, fonts etc.

*AC:*
Variables for:
- Size (sizes, margins, padding)
- Colors
- Borders
- Theme Families (violet vs teal)
- Fonts
- Elevation
- Shadows
- Timing
- Easing

Themes are created from a file of shared variables (e.g. a theme maps the color palette to intent: primary, success, warning, etc...).

PR #49, opened 2017-06-08
[Theme]: Add proper structure to styleVariables and mineralTheme

kylegach:
### Description

- Add proper structure to styleVariables and mineralTheme files
- Update Sample component and some examples to reflect new structure

Connects to #45 

### Motivation and context

We need Mineral UI's default theme to be real, so we can start building components.

### How has this been tested?

Manually confirmed that the Sample component renders correctly.

### Types of changes

### Checklist

Paint colors


brentertz:
The Sample component is not exported, thus this should not be a breaking change.  Also, it is only a color change, so again probably not a breaking change - though it does beg the question of how we define breaking changes for stylistic updates.  Lastly, we generally lowercase the "scope" portion of the commit message.

PR #63, opened 2017-06-14
Automatically document component props in the site

### LINKED ISSUE #34: [Site]: Investigate prop documentation options

coldpour:
### Description

Added a table to the site for each example that lists the props available to the component. The prop table is generated by react-docgen from the flow-types of the component's props. The table lists their type and a description of their use, whether they are required or their default value, and provides a text area where they can be edited by someone viewing the site.

### Motivation and context

closes #34 

### How has this been tested?

Ran the site and made sure that the props documented for each component were correct. Changed pages in the site and fixed a bug where the state wasn't getting updated properly which caused the 'Default' example to always have the first props loaded.

### Types of changes

### Checklist

not now chief, i'm in the zone


packages/site/src/components/PropList.js
```diff
@@ -0,0 +1,206 @@
...
+    <PropRow>
+      <PropCell><PropName>{name}</PropName></PropCell>
+      <PropCell>
+        <DefaultValue defaultValue={defaultValue} required={required} />
+      </PropCell>
+      <PropCell>
+        <PropP><PropType>{type}</PropType>&nbsp;{description}</PropP>
+      </PropCell>
+      {/* 1px allows textarea to use 100% height... magic */}
+      <PropCell style={{ height: '1px' }}>
```
coldpour: this doesn't work in FF :(

packages/site/src/components/PropList.js
```diff
@@ -0,0 +1,205 @@
...
+  type
+}: DocProps) {
+  return (
+    <PropRow>
+      <PropCell><PropName>{name}</PropName></PropCell>
+      <PropCell>
+        <DefaultValue defaultValue={defaultValue} required={required} />
+      </PropCell>
+      <PropCell>
+        <PropP><PropType>{type}</PropType>&nbsp;{description}</PropP>
```
brentertz: I believe that `type` should be a separate field from `description` and we should always attempt to document all props.  Separate columns improves the readability for me and it seems to be the most common pattern across various component libraries.

packages/site/src/components/PropList.js
```diff
@@ -0,0 +1,205 @@
...
+  name,
+  onPropChange,
+  required,
+  type
+}: DocProps) {
+  return (
+    <PropRow>
+      <PropCell><PropName>{name}</PropName></PropCell>
+      <PropCell>
+        <DefaultValue defaultValue={defaultValue} required={required} />
```
brentertz: As mentioned previously, I dislike combining the `default` and `required` fields.  Consider the following...

A string prop is required.  It is represented as having a default value of "required", which is likely not the actual default value, though that is what it makes me think.  This issue is also heightened for our visually impaired users as the "required" only differs based on color.  While I can certainly figure it out, I don't understand the desire to join these fields.

I would suggest to consider adding a small required `tag` or an `*` to the `prop` field.

Prior art:
* https://polaris.shopify.com/components/navigation/tabs#navigation
* http://www.material-ui.com/#/components/badge

PR #66, opened 2017-06-14
feat(style): Added 'createResetComponent' for styling components with…

limscoder:
### Description

Added `createResetComponent` function.
Has the same signature as 'createStyledComponent', but automatically includes reset styles.

Other discussed APIs:

* Always add reset: `createStyledComponent(...)` 
* Add reset only when option is present: `createdStyledComponent('div', {}, {applyReset: true})`
* Use `StyleReset` component: `createStyledComponent(StyleReset, ...)`. Requires tag to be passed at instantiation time: `<Root tag="button" />`.

### Motivation and context

Application of style reset was hard to grok.

### How has this been tested?

Tested with `Button` component.

### Types of changes

### Checklist

* [ ] [Accessibility](http://webaim.org/intro) and [inclusivity](https://24ways.org/2016/what-the-heck-is-inclusive-design/) considered
* [ ] Documentation created or updated

dog on beach


coldpour:
the name 'createResetComponent' initially confuses me. it makes me think that it's going to return a `<Reset />` component that I'm going to need to wrap around my component. I'm not sure what to do about that or what to suggest... just thought I should mention that in case anyone else felt similarly.

brentertz:
I have implemented an option within the `createStyledComponent` options param, called `includeStyleReset`, which defaults to false.  It is removed prior to passing the other options along to glamorous. This can be found on the `52-button` branch.

PR #73, opened 2017-06-16
feat(style): Show user-readable class names in dev mode.

limscoder:
### Description

Adds user-readable classnames in dev mode.

### DIFF
--- a/packages/style-utils/src/createStyledComponent.js
+++ b/packages/style-utils/src/createStyledComponent.js
@@ -15,7 +15,12 @@
  */
 
 /* @flow */
-import glamorous from 'glamorous';
+import glamorous from "glamorous";
+
+// $FlowFixMe
+glamorous.config.useDisplayNameInClassName = NODE_ENV === "development"
+  ? true
+  : false;
 
 export default function createStyledComponent(
   element: MnrlReactNode,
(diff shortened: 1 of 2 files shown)


packages/style-utils/src/createStyledComponent.js
```diff
@@ -15,7 +15,12 @@
  */
 
 /* @flow */
-import glamorous from 'glamorous';
+import glamorous from "glamorous";
+
+// $FlowFixMe
+glamorous.config.useDisplayNameInClassName = NODE_ENV === "development"
```
brentertz: I think this setting is something that should be left up to the consumer to configure in their own app.  I also believe that it will be the default in the future.  I experimented using it previously and really didn't feel like it added any value.

utils/makeWebpackConfig.js
```diff
@@ -13,20 +13,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const path = require('path');
-const webpack = require('webpack');
-const HtmlWebpackPlugin = require('html-webpack-plugin');
-const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
+const path = require("path");
```
brentertz: Why is your editor converting single quotes to double quotes?  We don't want that.  Do we maybe need an `.editorconfig` file?

brentertz:
Closing this issue as this is probably not the desired implementation.  I am still open to adding this configuration option to the site if you feel strongly.

PR #97, opened 2017-06-23
Speed up routing

coldpour:
### Description

change a linear time lookup to a constant time lookup by storing demos in an object where the slug is the key rather than in an array that needs to be searched.

### Motivation and context

I was reading the site code and wondering what the point of the DemoList component was since it was just rendering the ComponentDoc component and spreading the props. Then I saw that we were filtering our demo list every time we click a link in the nav. 

### How has this been tested?

went to all the pages in the site

### Types of changes

### Checklist


coldpour:
this should probably go right onto master and not wait for the button branch

coldpour:
we could also keep the slug in the demo objects and just create a keyed object for lookups... i just didn't think of that until just now....

packages/archetype/src/__demo__/index.js
```diff
@@ -19,11 +19,9 @@ import examples from './examples';
...
-    doc,
-    examples,
-    slug: 'archetype',
-    title: 'Archetype'
-  }
-];
+export default {
+  doc,
+  examples,
+  slug: 'archetype',
```
coldpour: remove the slugs

coldpour:
i think i broke github when i changed the base of this PR from the button branch to master...it's not showing the correct changes... so i'm going to make a new PR

Issue #52, opened 2017-06-08
[Button] initial component


coldpour:
As a Mineral UI developer, I want to develop a button component to validate our infrastructure and development experience.

Link to styleguide: https://ca2.invisionapp.com/share/9MBRILWTW

AC:

The button we build will handle Aaron's styleguide's specs for:

- Typography
    - currently looks like Open Sans (font file won't be provided yet)
    - system font fall back
    - need font-sizes to map to sizes
    - need weights
- Color
    - map theme colors to kinds & states
- Spacing
    - map to sizes?
- States
- Size / Shape
    - how many? 3?
- Variants (kinds) -- come up with a better word
- A11y
- Document any other relevant API updates

PR #109, opened 2017-06-27
[Button] Add Button component and more

brentertz:
### Description

* Add new `Button` package
* Refactor `mineralTheme`
* Rename `style-utils` package to `component-utils`
* Update tests suites to snapshot test all demo examples
* Misc related updates

### Motivation and context
`HelloWorld` components are nice, but we need to start building real components.  `Button` is our first real component.

connects #52

### How has this been tested?
* All examples are now being snapshot tested
* Manual visual testing using demo examples

### Types of changes

This PR includes a *New feature* and what would ordinarily be a *breaking change* (renaming the style-utils package), but since the previous state has not been published, the change is not being specified as "breaking".

### Checklist

button


limscoder:
It seems like a lot of these commits save new jest snapshots, I'm wondering how effective they will be in catching regressions.

brentertz:
@limscoder A large portion of the jest snapshot updates here are due to the fact that the mineral component examples are now wrapped with a `ThemeProvider` for testing.  In the browser, this is provided by the site itself.  While it is quite verbose here and will always be the case when adding new demo examples, smaller changes in the future should not be as large and should help to surface regressions.  That being said, visual-diff testing would also be beneficial.

Issue #53, opened 2017-06-08
[Button] Development Docs


coldpour:
As a consumer, I want development docs so that I know how to use it in my app.

AC:
The following docs are created:

1. Development Docs (props, code, examples)

Considerations:
* Reference the result of the site IA issue

kylegach:
- Remove How To Use & Changelog tabs
- Rename remaining tabs to be more clear
- Move Behavior to Design

PR #110, opened 2017-06-27
feat(site): move proptypes tables to top level of component page

coldpour:
### Description

move proptypes tables to top level of component page

### Motivation and context

the proptables used to be rendered per example, now they're only rendered once per component at the top of the page

connects #53 

### How has this been tested?

ran the demo

### Types of changes

refactor

### Checklist

hamburger

### DIFF
--- a/packages/site/src/components/ComponentDoc.js
+++ b/packages/site/src/components/ComponentDoc.js
@@ -20,6 +20,7 @@ import { createStyledComponent } from '@mineral-ui/component-utils';
 import ComponentDocExample from './ComponentDocExample';
 import Link from './Link';
 import styleReset from './styleReset';
+import PropTable from './PropTable';
 
 type Example = {
   component: MnrlReactComponent,
@@ -71,7 +72,6 @@ const styles = {
   h3: (props, theme) => ({
     margin: `${theme.measurement_d} 0 ${theme.measurement_c} 0`
   }),
-  h4: () => ({}),
   subnav: (props, theme) => ({
     borderBottom: `1px solid ${theme.color_gray}`,
     marginBottom: '2rem'
@@ -153,8 +153,7 @@ export default function ComponentDoc({
         <div>
           <H3>Behavior</H3>
           <p>{behavior}</p>
-          <H3>Props</H3>
-          <p>{"we'll move the prop table here"}</p>
+          {propDoc && renderPropDoc(propDoc)}
           {examples && renderExamples(examples, slug, propDoc)}
         </div>
         <H2 id="design">Design</H2>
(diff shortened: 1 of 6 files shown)


coldpour:
part of #53

limscoder:
We should give priority to getting the button PR in.

coldpour:
That's shouldn't conflict really

limscoder:
lgtm

packages/site/src/components/PropTable.js
```diff
@@ -148,64 +162,39 @@ function PropDoc({
...
 
-export default function PropList({
-  exampleProps = [],
-  onPropChange = () => {}
-}: Props) {
-  if (exampleProps.length < 1) {
-    return null;
-  }
+export default function PropTable({ exampleProps = {} }: Props) {
+  const foo = getExampleProps(exampleProps);
```
kylegach: foo?

packages/site/src/components/ComponentDoc.js
```diff
@@ -184,3 +183,7 @@ function renderExamples(
     </div>
   );
 }
+
+function renderPropDoc(propDoc: Object) {
+  return [<H3 key={0}>Props</H3>, <PropTable key={1} exampleProps={propDoc} />];
```
kylegach: Didn't realize we could return arrays of JSX elements. Will not miss the endless wrapping `div`s.

kylegach:
Please use the PR template as directed. If you'd like to modify the template, we can discuss that in a retro or via another PR with proposed changes.

packages/site/src/components/PropTable.js
brentertz: Agreed, foo?

brentertz:
As expressed several times in the past, I would really like to see the type field split out into its own column and not be arbitrarily joined with description.  It seems like a very minor change and at least tangentially associated to this PR.  Can it not be included here?  It would be nice to get this out prior to publishing (hopefully today).

coldpour:
@kylegach, I will update the PR description and follow up with my grievances of the template in a PR to that note

@brentertz, I didn't realize how strongly you felt on this point. Everytime, you brought it up, we'd raise the counterpoint of 3 columns being enough and saving us from empty columns, and then the conversation would stop. I think that should be a separate issue anyway since it is only tangentially related to this one and I'd rather handle separate concerns separately.

packages/site/src/components/PropTable.js
coldpour: oooh...shucks

brentertz:
@coldpour No worries about the type column.  We'll tackle it in a followup.

PR #118, opened 2017-06-28
[SITE] Add support for site routing on gh-pages

### LINKED ISSUE #28: [Site] Add routing
As a Mineral UI consumer, I would like a way to directly link to components on the demo site, so that I get to specific pages without needing to navigate from the home page. 

As a Mineral UI developer, I would like routing in the site, so that it is easier to manage, matches the design, has readable urls, is bookmarkable, etc.

**AC:**
- Route for each component
- Update component README's to point to permalink

**Other Considerations:**
- Site is currently deployed as an SPA
  - We either need to 
    - statically generate the files prior to deploying
    - use github 404 page to redirect 404 to root and use push state router
    - switch to Netlify - which supports both redirects and static deploy

brentertz:
### Description
Add support for site routing on gh-pages.  This should only be considered a temporary solution until we either prerender the site and/or switch to Netlify.

### Motivation and context

Our site package is currently deployed to Github pages as an SPA with client side routing.  With this naive approach, users must visit the root url and navigate using the in-app navigation alone.  They cannot navigate directly to a route, `components/button`, for example.  Github will give them a 404 page instead.  This PR introduces a hack that uses Github's support for a custom 404 page to capture the attempted URL, then redirect back to the root and use the saved url for routing the request.

closes #28 

### How has this been tested?

Verified locally - the 404 page is now used locally by webpack instead of just redirecting to /index.html.  While not fully necessary, and likely a tad slower, it allows us to validate the behavior locally.

I deployed the site and it is also working on mineral-ui.com.  You can try it too.  Click this link and if it takes you to the button docs, it worked.  http://mineral-ui.com/components/button

Tested cross browser.

### Types of changes

Is it a feature or a bug fix?  Not really sure.  I called it a bug fix.

### Checklist

hack

### DIFF
--- a/packages/site/public/404.html
+++ b/packages/site/public/404.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="x-ua-compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width">
+    <title>Mineral UI</title>
+    <script>
+      window.sessionStorage.redirect = window.location.href;
+    </script>
+    <meta http-equiv="refresh" content="0;URL='/'"></meta>
+  </head>
+  <body>
+    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+  </body>
+</html>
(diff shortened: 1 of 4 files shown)


packages/site/public/404.html
```diff
@@ -0,0 +1,17 @@
...
+    <meta http-equiv="x-ua-compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width">
+    <title>Mineral UI</title>
+    <script>
+      window.sessionStorage.redirect = window.location.href;
+    </script>
+    <meta http-equiv="refresh" content="0;URL='/'"></meta>
+  </head>
+  <body>
+    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
brentertz: FWIW, apparently, this file must be >512kb for IE to not ignore it - hence the `&nbsp;`s

PR #124, opened 2017-06-29
chore(scripts): stop `build` from running flow

coldpour:
if build stops the flow server, a subsequent `test` or `flow` will start it back up

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "bootstrap": "lerna bootstrap && npm run link-parent-bin",
-    "build": "npm run clean && npm run bootstrap && lerna run build && npm run flow:restart",
+    "build": "npm run clean && npm run bootstrap && lerna run build",
     "clean": "find packages -name dist -or -name node_modules -type d | xargs rm -rf",
     "commit": "git-cz",
     "commitmsg": "validate-commit-msg",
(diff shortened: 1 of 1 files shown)


brentertz:
Why would we want `build` to stop flow and not have it start thereafter?  We always want it running locally.  Currently flow is restarted in `build` in order to force it to pick up changes that it otherwise wouldn't notice.

coldpour:
I want it to stop during `build` and not start and run again because I rarely run `build` by itself. Usually, I run `build` followed by `test` or just `flow` if I'm focused on fixing flow errors. In that case, flow is running twice. It's also running twice in our ci for this same reason.

brentertz:
As a developer, I want flow running all the time locally, so that I become aware of errors as they occur and can fix them while I have context.  If I don't want to fix them at that time, I can also do it later, but at least I am aware of them.

Sometimes however, we make code changes that require a build in order for flow to pick them up.  Rather than have to run `build` and `flow:restart`, restarting flow was added to the build step.  This was done as convenience as I was having to execute the combination quite often.

If this disrupts your workflow, I am open to removing the `flow:restart` from the build step and going back to a 2 step approach.  If we went this route, I think we'd also likely need to update CI to start flow in order to be able to catch flow errors.

Regardless, I don't think that just stopping flow following a build is a good idea.

coldpour:
I had a few thoughts. 

first, atom with nuclide runs it's own separate flow server and runs `flow check` on file save. Kyle and I made sure of that by stoping flow in the terminal and then creating errors that atom kept picking up on. 

second, if we think of flow like a compiler, then perhaps `check` belongs in the build step and not in the test step. I would also be fine with that.

 but I stand by my assertion that running `check` twice in ci is confusing and unhelpful.

PR #133, opened 2017-07-06
fix(button): export should allow default import

coldpour:
### Description
make Button a default export for easy importing, consistent with our other components, and consistent with the Readme.

### Motivation and context

I pulled our button into a create-react-app and it errored a bunch until I destructured the import.

### How has this been tested?

This has not been directly tested, but it follows the pattern of our other components, hello and world, that are allowed to use the default export.

### Types of changes

- Bug fix: BREAKING

### DIFF
--- a/packages/button/src/index.js
+++ b/packages/button/src/index.js
@@ -15,4 +15,4 @@
  */
 
 /* @flow */
-export { default as Button } from './Button';
+export { default } from './Button';


brentertz:
Yep, good catch.  I think this was missed when we went back from having multiple Button components to a single one.

PR #170, opened 2017-07-14
[Site] Add capability of rendering multiple demos per package

kylegach:
### Description
- A package can now export either a single object or an array of objects
for its demo(s)
- Subcomponents are indented in the nav via a flag in the demo object
- Add utils dir to Site

### Motivation and context

Some of our components, like the soon-to-come Card, have subcomponents in the same package. We need a way to display these subcomponents in the nav and display their demos without requiring a separate package.

screen shot 2017-07-14 at 11 53 24 am

### How has this been tested?

These changes were developed along with the Card component (see screenshot above) and then extracted from that branch into their own PR for clarity. In addition to confirming it works while developing, I also arbitrarily wrapped Button's demo in an array to test the functionality after I extracted it. The following scenarios work as expected:

- `cd packages/site`, `npm run start`
- `cd packages/button`, `npm run start` (package with a single demo)
- `cd packages/card`, `npm run start` (package with multiple demos)

### Types of changes

### Checklist

Submarine


packages/site/src/components/Nav.js
```diff
@@ -62,7 +64,7 @@ export default function Nav({ className, demos }: Props) {
   const demoLinks = Object.keys(demos).map(slug => {
     const demo = demos[slug];
     return (
-      <ListItem key={slug}>
+      <ListItem key={slug} indent={demo.subcomponent}>
```
brentertz: Consider renaming prop `indent` to `subcomponent` of `isSubcomponent`.  This way, you are not "limited" to indentation styles if the design changes later.

packages/site/src/components/App.js
```diff
@@ -60,27 +61,40 @@ const Root = createStyledComponent('div', styles.app);
...
+const getFirstDemo = (demos: Object): string => {
+  for (const first in demos) {
+    return first.toString();
+  }
+  return '';
+};
+
 export default function App({ className, demos }: Props) {
-  if (demos.slug) {
+  let siteDemos = {};
```
brentertz: This logic feels like it is getting a bit intertwined.  Is it possible to optimize this for improved clarity?  e.g.

```js
if (demos.slug) {
  return <ComponentDoc {...demos} />;
}

const siteDemos = Array.isArray(demos) ? pullOutKeys(demos, 'slug') : demos;
```

packages/site/src/components/App.js
```diff
@@ -60,27 +61,40 @@ const Root = createStyledComponent('div', styles.app);
 const Nav = createStyledComponent(_Nav, styles.nav);
 const Main = createStyledComponent('main', styles.main);
 
+const getFirstDemo = (demos: Object): string => {
```
brentertz: Consider renaming to `getDefaultDemo`.

brentertz: Simpler implementation?

```js
const getDefaultDemo = (demos: Object): string => Object.keys(demos)[0];
```

packages/site/src/components/App.js
kylegach: Flow yells at you because `demos` in that first line can be an object or an array. This works, though:

```
if (!Array.isArray(demos) && demos.slug) {
  return <ComponentDoc {...demos} />;
}

const siteDemos = Array.isArray(demos) ? pullOutKeys(demos, 'slug') : demos;
```

Still like that better than what's currently there?

kylegach:
All comments addressed.

Issue #142, opened 2017-07-11
[Card] initial component


Mikewaite:
As a consumer, I want a Card component, so that I can lay out content that belongs in a card.

AC:
1. Typography
2. Color
3. Spacing
6. States
8. Variants
9. A11y

Other Considerations:
- Consider basic Card content, header, image, and actions (footer) separately, but in a cohesive manner

PR #171, opened 2017-07-14
[Card] Initial component

kylegach:
### Description

- Add Card, CardActions, CardBlock, CardImage, and CardTitle components (and also a non-exported CardRow)
- Add demos for each of the above
- Add `color_black`, `color_white`, and `spacing_triple` to mineralTheme

Connects #142

### Motivation and context

All the cool kids have a Card component.

screen shot 2017-07-14 at 4 54 33 pm

### How has this been tested?

Checked demo in all supported browsers.

### Types of changes

### Checklist

Shuffling cards


packages/card/README.md
```diff
@@ -0,0 +1,35 @@
...
+  ```sh
+  npm install --save @mineral-ui/card
+  ```
+
+
+## Usage
+
+```jsx
+import React from 'react';
+import { Card, CardTitle, CardBlock } from '@mineral-ui/card';
```
kylegach: I wasn't sure if Card should be a default export or not. This would be our first component without a default.

packages/card/src/__demo__/examples/card.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+/* @flow */
+import React from 'react';
+import { createStyledComponent } from '@mineral-ui/component-utils';
+import Card from '../../Card';
+import CardBlock from '../../CardBlock';
+import CardTitle from '../../CardTitle';
+
+const Root = createStyledComponent('div', {
```
kylegach: Not sure that we should keep all of these in here, as Cards automatically stretch to fill the available width and this makes it seem otherwise.

packages/card/src/__tests__/__snapshots__/Archetype.spec.js.snap
```diff
@@ -0,0 +1,143 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
kylegach: Oops. I'll delete this.

kylegach:
Please note that this is built off of the branch for #170.

packages/card/src/__tests__/Card.spec.js
```diff
@@ -0,0 +1,48 @@
...
+import { shallow } from 'enzyme';
+import Card from '../Card';
+import cardExamples from '../__demo__/examples/cardIndex';
+import cardActionsExamples from '../__demo__/examples/cardActionsIndex';
+import cardBlockExamples from '../__demo__/examples/cardBlockIndex';
+import cardImageExamples from '../__demo__/examples/cardImageIndex';
+import cardTitleExamples from '../__demo__/examples/cardTitleIndex';
+import testDemoExamples from '../../../../utils/test/testDemoExamples';
+
+const examples = [].concat(
```
kylegach: This works, but would it better to refactor `testDemoExamples` to accept multiple example arguments?

packages/component-utils/src/getNormalizedValue.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+export default function getNormalizedValue(value: string, base: string) {
```
kylegach: Open to suggestions on naming this thing.

packages/site/src/utils/pullOutKeys.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+export default function pullOutKeys(arr: Array<Object>, key: string) {
```
kylegach: Open to suggestions on naming this thing.

packages/site/src/utils/flatten.js
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+export default function flatten(arr: Array<*>) {
+  return arr.reduce(function(flat, toFlatten) {
+    return flat.concat(
+      Array.isArray(toFlatten) ? flatten(toFlatten) : toFlatten
```
coldpour: i think we should probably have a test for this util. i also think it will break for things that are too deeply nested. JS doesn't have the tail-recursion optimization and so this sort of operation can stack overflow. while recursion is pretty, it's best in JS to do things imperatively with loops.

packages/site/package.json
```diff
@@ -20,6 +20,7 @@
   },
   "devDependencies": {
     "@mineral-ui/button": "^1.0.0",
+    "@mineral-ui/card": "^0.0.1",
```
brentertz: New packages should start out with version 0.0.0.  They will be updated during the first publish.

packages/card/src/__demo__/index.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+/* @flow */
+import cardExamples from './examples/cardIndex';
+import cardActionsExamples from './examples/cardActionsIndex';
+import cardBlockExamples from './examples/cardBlockIndex';
+import cardImageExamples from './examples/cardImageIndex';
+import cardTitleExamples from './examples/cardTitleIndex';
+
+const cardDoc = require('!!react-docgen-loader!../Card.js');
```
brentertz: I am getting a bunch of `docgen` errors on the Card components when running this branch locally.  

Resultantly the documentation does not show up properly in the site.   It must have something to do with the component type.  Perhaps docgen expects a particular format?  Note that `CardTitle` works, but the others do not.

```sh
docgen-loader Error: No suitable component definition found.
```

packages/card/src/CardActions.js
```diff
@@ -0,0 +1,43 @@
...
+      paddingRight: theme.CardActions_padding,
+
+      '&:last-child': {
+        marginBottom: 0,
+        marginTop: theme.CardActions_marginTop
+      }
+    };
+  },
+  {
+    displayName: 'CardActions'
```
brentertz: I am assuming that you have thought this through and that the answer is no, but I am curious as to why we don't have `includeStyleReset: true` on the subcomponents.

packages/component-utils/src/getNormalizedValue.js
brentertz: Maybe just add a comment describing what it does or why it is needed.

packages/card/src/__demo__/examples/arbitraryChildren.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+}));
+
+function Example() {
+  return (
+    <Root>
+      <Card>
+        <CardTitle>Card title</CardTitle>
+        <CustomContent>
+          <Button fullWidth onClick={() => {}}>Button</Button>
```
brentertz: `onClick` is not a required prop of `Button`, so this handler is unnecessary.

packages/card/src/__demo__/index.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import cardExamples from './examples/cardIndex';
```
brentertz: It is difficult at a glance to identify which examples go with which components.  Using folders might help organization a bit.  I doubt there is a good reason to keep the examples directory flat. 

```
__demo__
    index.js
    examples/
        index.js
        card/
             index.js
            …
        cardAction/
            index.js
            …
        cardBlock/
            index.js
            …
        cardImage/
            index.js
            …
        cardTitle/
            index.js
            …
```

packages/card/src/__tests__/Card.spec.js
brentertz: How about just add an `index.js` to the examples directory like other components, then import that here?

packages/component-utils/src/styleVariables.js
```diff
@@ -15,6 +15,11 @@
  */
 
 /* @flow */
+export const blackAndWhite = {
```
brentertz: Following how we handled the rest of the palette.   It seems like `black` and `white` should just be keys inside the `color` object in `styleVariables`.  Then, in `mineralTheme`, we should define `color_black` and `color_white` using those keys.

packages/card/src/__demo__/examples/crazyOrder.js
```diff
@@ -0,0 +1,153 @@
...
+  '& > *:not(:first-child)': {
+    marginLeft: '2em'
+  }
+});
+
+function Example() {
+  return (
+    <Root>
+      <Card>
+        <CardImage src="http://fillmurray.com/1000/563" alt="Bill Murray" />
```
brentertz: Wow, fillmurray.com is pretty awesome.  It really adds a touch of class to your examples.  We should however have a follow-up story to replace these images with mineral related placeholders once we get some design assets.

packages/card/README.md
kylegach: > I am suggesting this because everything is still coming from the@mineral-ui/card package and it makes my brain think that the default would be a Card. I think this would be a good pattern for us to follow.

Agreed.

packages/site/src/utils/flatten.js
kylegach: This is better discussed in #170, but I opted to simply import lodash's flatten (which only supports one level, but there are other packages to support recursion if it turns out we need that) instead of writing our own.

packages/card/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@mineral-ui/card",
+  "version": "0.3.2",
```
kylegach: 🤦‍♂️ Thanks.

packages/card/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "description": "Card component",
+  "license": "Apache-2.0",
+  "private": true,
```
kylegach: 🤦‍♂️ Thanks.

packages/card/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "devDependencies": {
+    "@mineral-ui/component-utils": "^0.1.2"
+  },
+  "peerDependencies": {
+    "@mineral-ui/component-utils": "^0.1.2",
+    "glamor": "^2.0.0",
+    "glamorous": "^3.0.0",
+    "react": "^15.0.0",
+    "react-dom": "^15.0.0"
+  },
```
kylegach: 🤦‍♂️ Thanks.

packages/card/src/__demo__/index.js
kylegach: 🤦‍♂️ I'm embarrassed I didn't think of that. Thanks for the suggestion.

packages/card/src/__demo__/index.js
kylegach: Aww, bummer. I really liked the elegance of directly exporting the output of `createStyledComponent`, but you're right, it totally breaks docgen-loader.

packages/card/src/CardActions.js
kylegach: All of the subcomponents are only intended to be used within the root Card component. All of the reset styles, with the exception of `outline`, inherit... so we should be fine as-is.

packages/component-utils/src/styleVariables.js
kylegach: Yeah, that probably is better. I let DRY be a little too important, there.

packages/card/src/__tests__/Card.spec.js
kylegach: I like that idea, Brent.

Issue #29, opened 2017-05-15
[Site] Code example


brentertz:
As a Mineral UI consumer, I want code examples in the site, so that I can quickly understand how to use the components.

**AC:**
- Syntax highlighting
- Code matches prop demo example

**Other considerations:**
- Live editable code
  - https://github.com/philpl/react-live
  - https://formidable.com/open-source/component-playground
  - http://codemirror.net/

PR #175, opened 2017-07-17
Site/code example

coldpour:
### Description

This PR adds a live code editor to each example in the site. The code editor is from React-Live; it takes a string of source code, compiles it with React and any other objects you put in scope and renders it in the Live Preview area. Examples that wish to hide their source can do so by specifying `hideSource: true` and examples which need a special color for their background can specify a `backgroundColor: '#ccc'` or other color string.

This removes the need to have both the source string and the example Component in the example file. Development of examples can be done by specifying scope objects and an initial bit of code and then coding right in the LiveEditor on the site.

### Motivation and context

This makes several of our examples on the site more helpful. A prime example is the Component-Utils package which before just showed the same color box over and over. Now it shows the code behind that colored box and how the API can be used in different ways to achieve the same result.

...i guess i could record something, but you should really just try it out for yourself...

### How has this been tested?

running the demo and updating the snapshots. The `testDemoExamples` function needed to be updated to use the LiveProvider and pass in the source string and scope so that they kept rendering without the duplicate JSX from `example.component`.

### Types of changes

### Checklist

* [ ] Documentation created or updated

shocked


packages/archetype/src/__tests__/__snapshots__/Archetype.spec.js.snap
```diff
@@ -125,19 +125,145 @@ exports[`Archetype Default 1`] = `
...
-        <Archetype>
-          <glamorous(div)>
-            <div
-              className="css-nil"
-            >
-              Archetype
-            </div>
-          </glamorous(div)>
-        </Archetype>
+    <LiveProvider
```
kylegach: It's not awesome that all of our snapshots will no longer match if we change our LivePreview styles. Wonder if there's a way around that.

coldpour: LivePreview is needed to render the string of source, but there are no styles attached to the snapshot tests, see `testDemoExamples.js`. We're just sending the source string into an unstyled LiveProvider and LivePreview so that they get rendered. 

The styles for the demo are in ComponentDocExample and are separate. You can modify those styles without affecting the snapshot tests. 

What will affect the snapshot tests is if you modify the styles on the demo examples, for instance on the layout components wrapping several buttons in a single example.

packages/button/src/__demo__/examples/danger.js
```diff
@@ -15,33 +15,22 @@
  */
 
 /* @flow */
-import React from 'react';
 import { createStyledComponent } from '@mineral-ui/component-utils';
 import Button from '../../Button';
 
-const Root = createStyledComponent('div', {
+export const Root = createStyledComponent('div', {
```
coldpour: ...idk

riley:
still does weird stuff when I put in emojis, but it's probably not your implementation 🤔

coldpour:
@riley try `<World text={'\ud83d\ude03'} />`

utils/test/testDemoExamples.js
```diff
@@ -19,11 +19,13 @@ import React from 'react';
 import { mount } from 'enzyme';
 // $FlowFixMe
 import { ThemeProvider } from '@mineral-ui/component-utils';
+// $FlowFixMe
```
brentertz: You can get rid of these `$FlowFixMe`'s for `react-live`.  You just need to create the stub using `flow-typed`.

```sh
./node_modules/.bin/flow-typed create-stub react-live@latest
npm run flow:restart
```

packages/site/src/components/ComponentDocExample.js
```diff
@@ -27,113 +29,65 @@ const styles = {
...
       <Root className={className}>
         <H4>{title}</H4>
-        {typeof description === 'string'
-          ? <Graf>{description}</Graf>
-          : description}
-        <Resizable>
-          <Component {...componentProps} />
-        </Resizable>
+        {typeof description === 'string' ? <P>{description}</P> : description}
+        <LiveProvider code={source} scope={scope}>
```
brentertz: The `<LiveProvider />` is injecting the same bunch of `prism` CSS for each example.  We should ideally set `mountStylesheet={false}` and then load the [react-live.css](https://github.com/FormidableLabs/react-live/blob/master/react-live.css) once for the whole site.  This should also have the effect of removing it from the snapshots.

This could be done as a followup story if you are anxious to get this deployed.

packages/archetype/src/__tests__/__snapshots__/Archetype.spec.js.snap
brentertz: Using mountStylesheet={false} as mentioned elsewhere would likely help here.

brentertz:
I am hopeful that we can figure out a way to get syntax highlighting of the code inside of the template strings.

coldpour:
connects #29

PR #211, opened 2017-07-28
WIP: [Link] component

### LINKED ISSUE #149: [Link] component
As a consumer using react router, i want my links to go through my routing and have all the CA styles. Additionally, for people that know the difference between the usage of link and button, we should provide a solution for a link, if not a link component.

AC:
1. Provide a consistently styled link.
2. Make sure this link works with React Router.
3. Show your work

Other considerations:
- Warn if using href on Button or onClick on Link?
- Do and Don'ts to reflect usage as well.
- Should this be its own component?
    - Or should Button take an href prop?
- Does Button need a way to display without padding?

riley:
### Description
Added a Link component. This really just a wrapper around an `<a>` since we don't want to re-invent the wheel re: keyboard events, a11y, etc.

closes #149 

### Motivation and context
We decided that Link needs to be a separate component than Button, since a button is a form element, while an anchor is not.

### How to test
1. `npm run build`
2. navigate to Site > Link
3. Play with the links and make sure all the normal props are passed thru
4. It's pretty much just an anchor tag

### Types of changes
- New feature

### Checklist

hypertext is awesome

### DIFF
--- a/packages/link/README.md
+++ b/packages/link/README.md
@@ -1,0 +1,26 @@
+# Link
+
+Mineral UI `Link` component. Find out more about [Mineral UI](https://github.com/mineral-ui/mineral-ui).
+
+
+## Installation
+
+1. Install project [peer dependencies](../../docs/peer-dependencies.md), if not already installed
+
+2. Install the package
+
+  ```sh
+  npm install --save @mineral-ui/link
+  ```
+
+
+## Usage
+
+```jsx
+import React from 'react';
+import Link from '@mineral-ui/link';
+
+export default function MyComponent() {
+  return <p>Check out <Link href="https://mineral-ui.com">Mineral UI</Link>!</p>;
+}
+```
--- a/packages/link/package.json
+++ b/packages/link/package.json
@@ -1,12 +1,12 @@
 {
-  "name": "@mineral-ui/site",
-  "version": "0.9.0",
-  "description": "Site",
+  "name": "@mineral-ui/link",
+  "version": "0.0.0",
+  "description": "Link component",
   "license": "Apache-2.0",
-  "private": true,
-  "main": "dist/index.js",
+  "main": "dist/umd/index.js",
+  "module": "dist/es/index.js",
   "scripts": {
-    "build": "TARGET=site $npm_package_config_script_path/build.sh",
+    "build": "$npm_package_config_script_path/build.sh",
     "build:analyze": "$npm_package_config_script_path/build-analyze.sh",
     "clean": "$npm_package_config_script_path/clean.sh",
     "commit": "$npm_package_config_script_path/commit.sh",
@@ -15,27 +15,21 @@
     "jest": "$npm_package_config_script_path/jest.sh",
     "lint": "$npm_package_config_script_path/lint.sh",
     "pack-ls": "$npm_package_config_script_path/pack-ls.sh",
-    "start": "TARGET=site webpack-dev-server --progress",
+    "start": "DEMO_LIST_PATH=\"link\" $npm_package_config_script_path/start.sh",
     "test": "$npm_package_config_script_path/test.sh"
   },
   "devDependencies": {
-    "@mineral-ui/button": "^2.0.0",
-    "@mineral-ui/card": "^1.0.0",
-    "@mineral-ui/component-utils": "^1.0.0",
-    "@mineral-ui/icon": "^1.0.0",
-    "@mineral-ui/link": "^0.0.0"
+    "@mineral-ui/component-utils": "^1.0.0"
   },
   "peerDependencies": {
-    "@mineral-ui/button": "^2.0.0",
-    "@mineral-ui/card": "^1.0.0",
     "@mineral-ui/component-utils": "^1.0.0",
-    "@mineral-ui/icon": "^1.0.0",
-    "@mineral-ui/link": "^0.0.0",
     "glamor": "^2.0.0",
     "glamorous": "^3.0.0",
     "react": "^15.0.0",
-    "react-dom": "^15.0.0",
-    "react-live": "1.7.0"
+    "react-dom": "^15.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
   },
   "config": {
     "commitizen": {
--- a/packages/link/src/Link.js
+++ b/packages/link/src/Link.js
@@ -1,0 +1,73 @@
+/**
+ * Copyright 2017 CA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import React from 'react';
+import { createStyledComponent } from '@mineral-ui/component-utils';
+
+type Props = {
+  /** Content of the Link */
+  children: MnrlReactNode,
+  /** A URL or a URL fragment that the link points to */
+  href?: string,
+  /** Element to be used as the root node - e.g. "a" or { ReactRouterLink } */
+  element?: $FlowFixMe // Should allow string | React class
+};
+
+const linkStyles = (props, baseTheme) => {
+  const theme = {
+    Link_color: baseTheme.color_link,
+    Link_color_hover: baseTheme.color_link_hover,
+    Link_color_focus: baseTheme.color_link_focus,
+    Link_color_active: baseTheme.color_link_active,
+    Link_fontSize: `${2 * baseTheme.fontSize_base}px`,
+    Link_fontWeight: baseTheme.fontWeight_regular,
+
+    ...baseTheme
+  };
+
+  return {
+    color: theme.Link_color,
+    fontWeight: theme.Link_fontWeight,
+    fontSize: theme.Link_fontSize,
+    outline: null, // Using the browser's focus styles until design is finalized
+
+    '&:hover': {
+      color: theme.Link_color_hover
+    },
+    '&:focus': {
+      color: theme.Link_color_focus
+    },
+    // `:active` must be last, to follow LVHFA order:
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/:active
+    '&:active': {
+      color: theme.Link_color_active
+    }
+  };
+};
+
+/**
+ * The Link component creates a hyperlink to another URL.
+ */
+export default function Link({ children, element = 'a', ...restProps }: Props) {
+  const Root = createStyledComponent(element, linkStyles, {
+    displayName: 'Link',
+    includeStyleReset: true,
+    rootEl: element
+  });
+
+  return <Root {...restProps}>{children}</Root>;
+}
--- a/packages/link/src/__demo__/examples/index.js
+++ b/packages/link/src/__demo__/examples/index.js
@@ -15,14 +15,9 @@
  */
 
 /* @flow */
-import flatten from 'lodash/flatten';
-import createKeyMap from '../utils/createKeyMap';
-import button from '../../../button/src/__demo__';
-import card from '../../../card/src/__demo__';
-import componentUtils from '../../../component-utils/src/__demo__';
-import icon from '../../../icon/src/__demo__';
-import link from '../../../link/src/__demo__';
+import link from './link';
+import children from './children';
+import reactRouter from './reactRouter';
+import states from './states';
 
-const demos = flatten([button, card, icon, link, componentUtils]);
-
-export default createKeyMap(demos, 'slug');
+export default [link, children, reactRouter, states];
--- a/packages/link/src/__demo__/examples/link.js
+++ b/packages/link/src/__demo__/examples/link.js
@@ -15,14 +15,16 @@
  */
 
 /* @flow */
-import flatten from 'lodash/flatten';
-import createKeyMap from '../utils/createKeyMap';
-import button from '../../../button/src/__demo__';
-import card from '../../../card/src/__demo__';
-import componentUtils from '../../../component-utils/src/__demo__';
-import icon from '../../../icon/src/__demo__';
-import link from '../../../link/src/__demo__';
+import Link from '../../Link';
 
-const demos = flatten([button, card, icon, link, componentUtils]);
-
-export default createKeyMap(demos, 'slug');
+export default {
+  title: 'Basic link',
+  description: 'Links used in prose create hypertext.',
+  scope: { Link },
+  source: `<p>
+  <Link href="http://example.com">Light years</Link> star stuff
+  <Link href="http://example.com">harvesting</Link> star light citizens of distant
+  <Link href="http://example.com">epochs</Link> encyclopaedia galactica vastness is bearable only through love,
+  shores of the cosmic ocean!
+</p>`
+};
--- a/packages/link/src/__demo__/examples/reactRouter.js
+++ b/packages/link/src/__demo__/examples/reactRouter.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2017 CA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import Link from '../../Link';
+import {
+  BrowserRouter,
+  Link as ReactRouterLink,
+  Redirect,
+  Route,
+  Switch
+} from 'react-router-dom';
+
+export default {
+  title: 'ReactRouter',
+  description: `This example passes a ReactRouter Link to a Mineral UI Link's element prop in order to apply Mineral UI styles and achieve client side routing.`,
+  scope: { BrowserRouter, Link, ReactRouterLink, Redirect, Route, Switch },
+  source: `<BrowserRouter>
+  <Switch>
+    <Route
+      path="/components/:componentId/:path?"
+      render={route => {
+        const path = route.match.params.path;
+
+        if ('page-2' === path) {
+          return <Link to="/components/link/page-1" element={ReactRouterLink}>Page 1</Link>;
+        }
+        return <Link to="/components/link/page-2" element={ReactRouterLink}>Page 2</Link>;
+      }}
+    />
+    <Redirect from="/" to="/components/link" />
+  </Switch>
+</BrowserRouter>`
+};
--- a/packages/link/src/__demo__/examples/states.js
+++ b/packages/link/src/__demo__/examples/states.js
@@ -13,27 +13,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import darken from 'polished/lib/color/darken';
-import { Link } from 'react-router-dom';
-import { createStyledComponent } from '@mineral-ui/component-utils';
-import styleReset from './styleReset';
 
-const styles = (props, theme) => ({
-  ...styleReset(theme),
-  textDecoration: 'none',
-  ':link,:visited': {
-    color: theme.color_interactive
-  },
-  ':hover,:focus': {
-    color: darken(0.1, theme.color_interactive)
-  },
-  ':active': {
-    color: darken(0.15, theme.color_interactive)
+/* @flow */
+import { simulate } from 'glamor';
+import { createStyledComponent } from '@mineral-ui/component-utils';
+import Link from '../../Link';
+
+const DemoLayout = createStyledComponent('div', {
+  '& > a': {
+    marginRight: '0.5rem'
   }
 });
 
-export const RouterLink = createStyledComponent(Link, styles, {
-  forwardProps: ['to']
-});
-
-export default createStyledComponent('a', styles);
+export default {
+  title: 'States',
+  hideSource: true,
+  scope: { Link, DemoLayout, simulate },
+  source: `<DemoLayout>
+  <Link href="http://example.com">Default</Link>
+  <Link {...simulate('hover')} href="http://example.com">Hover</Link>
+  <Link {...simulate('focus')} href="http://example.com">Focus</Link>
+  <Link {...simulate('active')} href="http://example.com">Active</Link>
+</DemoLayout>`
+};
--- a/packages/link/src/__tests__/Link.spec.js
+++ b/packages/link/src/__tests__/Link.spec.js
@@ -15,14 +15,25 @@
  */
 
 /* @flow */
-import flatten from 'lodash/flatten';
-import createKeyMap from '../utils/createKeyMap';
-import button from '../../../button/src/__demo__';
-import card from '../../../card/src/__demo__';
-import componentUtils from '../../../component-utils/src/__demo__';
-import icon from '../../../icon/src/__demo__';
-import link from '../../../link/src/__demo__';
+import React from 'react';
+import { shallow } from 'enzyme';
+import Link from '../Link';
+import examples from '../__demo__/examples';
+import testDemoExamples from '../../../../utils/test/testDemoExamples';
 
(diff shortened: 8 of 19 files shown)


packages/link/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Usage
+
+```jsx
+import React from 'react';
+import Link from '@mineral-ui/link';
+
+export default function MyComponent() {
+  return (
+    <div>
+      <Link href="https://yourawesomesite.com">
```
kylegach: We should link to `http://mineral-ui.com`.

packages/link/src/Link.js
```diff
@@ -0,0 +1,77 @@
...
+type Props = {
+  /** Content of the links. Can be anything, but see LinkBlock, LinkImage, and LinkTitle. */
+  children: MnrlReactNode,
+  /** URI tied to the Link */
+  href: string
+};
+
+const Root = createStyledComponent(
+  'a',
+  (props, baseTheme) => {
```
kylegach: Lines 29 – 31 should all be on one line. Not sure why Prettier didn't catch that.

packages/link/src/Link.js
```diff
@@ -0,0 +1,77 @@
...
+  return {
+    fontSize: theme.Link_fontSize
+  };
+});
+
+/**
+ * Link component
+ */
+export default function Link(props: Props) {
+  const rootProps = {
```
kylegach: You need to pull `children` out so that you're not applying it to `Root`:

```
const rootProps = {
    children,
    ...restProps
};
return <Root {...restProps}>...
```

packages/link/src/__demo__/examples/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import linkExamples from './link';
+
+const examples = [].concat(linkExamples);
```
kylegach: This pattern (and the entire file) is only necessary for packages with subcomponents.

packages/link/src/__demo__/examples/link/arbitraryChildren.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/* @flow */
+import { mineralTheme } from '@mineral-ui/component-utils';
+import Link from '../../../Link';
+
+export default {
+  title: 'Arbitrary children',
+  // $FlowFixMe
+  backgroundColor: mineralTheme.color_gray_10,
```
kylegach: Probably not necessary.

packages/link/src/__demo__/examples/link/states.js
```diff
@@ -0,0 +1,38 @@
...
+    marginRight: '0.5rem'
+  }
+});
+
+export default {
+  title: 'States',
+  hideSource: true,
+  scope: { Link, DemoLayout, simulate },
+  source: `<DemoLayout>
+  <Link href="http://example.com">Don't Panic</Link>
```
kylegach: Love it. :)

packages/link/src/__tests__/Link.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+describe('Link', () => {
+  let link;
+  it('renders', () => {
+    link = renderLink({ href: 'http://example.com' }, 'Children');
+
+    expect(link.exists()).toEqual(true);
+  });
+
+  describe('onClick', () => {
```
kylegach: This is just default browser behavior. Since we're not doing anything with the `onClick` prop in the component (it just gets passed through, because it's a standard DOM attribute for `a`), we probably don't need tests for it.

packages/link/src/linkTheme.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+export default function linkTheme(baseTheme: Object) {
```
kylegach: I think we decided to keep these in the component file, unless there's plenty of reason (like subcomponents) to split it out.

packages/link/README.md
```diff
@@ -0,0 +1,32 @@
+# Card
```
coldpour: needs updates

packages/link/README.md
coldpour: yup

packages/link/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "devDependencies": {
+    "@mineral-ui/component-utils": "^0.3.0"
+  },
+  "peerDependencies": {
+    "@mineral-ui/component-utils": "^0.3.0",
+    "glamor": "^2.0.0",
+    "glamorous": "^3.0.0",
+    "react": "^15.0.0",
+    "react-dom": "^15.0.0",
+    "react-router-dom": "^4.1.2"
```
coldpour: @brentertz is this cool? hot or not?

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import { Link as RouterLink } from 'react-router-dom';
+import { createStyledComponent } from '@mineral-ui/component-utils';
+
+type Props = {
+  /** Content of the link. Can be anything */
+  children: MnrlReactNode,
+  /** URI tied to the Link */
+  to: string
+};
```
coldpour: something is messed up with our props cause the proptable is not rendering in the site...

packages/link/src/__demo__/examples/link.js
```diff
@@ -0,0 +1,26 @@
...
+/* @flow */
+import { mineralTheme } from '@mineral-ui/component-utils';
+import Link from '../../Link';
+
+export default {
+  title: 'Basic link',
+  description: 'an anchor tag',
+  scope: { Link },
+  source: `<Link to="https://example.com">Make it so</Link>`
+};
```
coldpour: need a prose example

packages/site/src/components/__tests__/__snapshots__/App.spec.js.snap
```diff
@@ -912,6 +912,78 @@ exports[`App renders correctly 1`] = `
...
+                  "World": [Function],
+                },
+                "source": "<World text=\\"Felix\\" />",
+                "title": "Custom Text",
+              },
+            ],
+            "slug": "world",
+            "title": "World",
+          },
+>>>>>>> WIP: linting
```
coldpour: oops

packages/link/CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
+# Change Log
```
brentertz: Please delete all content within this file.

brentertz:
This PR has a `packages/link/package-lock.json` file which must be deleted.  We are not using lock files.  If it reappears, it is a smell that you have some mineral dependency versioning problems on your branch.  I can help.

packages/link/src/__demo__/examples/states.js
```diff
@@ -13,27 +13,26 @@
...
-  forwardProps: ['to']
-});
-
-export default createStyledComponent('a', styles);
+export default {
+  title: 'States',
+  hideSource: true,
+  scope: { Link, DemoLayout, simulate },
+  source: `<DemoLayout>
+  <Link to="http://example.com">Don't Panic</Link>
```
brentertz: As Kyle suggested, linking to mineral-ui would be better.

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+  const units = measurement.replace(value, '');
+  return `${func(value)}${units}`;
+}
+
+const Root = createStyledComponent(RouterLink,
+  (props, theme) => {
+    return {
+      color: theme.color_link,
+      fontWeight: theme.Link_fontWeight || theme.fontWeight_regular,
+      fontSize: theme.Link_fontSize || transform(theme.fontSize_base, v => v*2),
```
brentertz: I'd like to get @kylegach to review these styles.

packages/site/src/components/Footer.js
```diff
@@ -17,7 +17,7 @@
 /* @flow */
 import React from 'react';
 import { createStyledComponent } from '@mineral-ui/component-utils';
-import Link from './Link';
+import Link from '@mineral-ui/link';
```
brentertz: Good thinking to update the site to use the new Link component.  Let's just try to keep this stuff as a separate commit.

packages/link/src/__demo__/examples/arbitraryChildren.js
```diff
@@ -0,0 +1,28 @@
...
+
+/* @flow */
+import { mineralTheme } from '@mineral-ui/component-utils';
+import Link from '../../Link';
+
+export default {
+  title: 'Arbitrary children',
+  description: 'A Link will render any children.',
+  scope: { Link },
+  source: `<Link to="http://example.com">
```
brentertz: This link doesn't actually work.  I click it and it changes the url to `http://0.0.0.0:8080/http://example.com` but it does not navigate.

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+  {
+    displayName: 'Link',
+    includeStyleReset: true
+  }
+);
+
+/**
+ * Link component
+ */
+export default function Link({children, ...otherProps}: Props) {
```
brentertz: Our convention is to use the name `restProps` rather than `otherProps`.

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+    };
+  },
+  {
+    displayName: 'Link',
+    includeStyleReset: true
+  }
+);
+
+/**
+ * Link component
```
brentertz: Perhaps we can find a better way to describe this component, rather than simply Link component.  Check here for some inspiration.  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

packages/link/src/__demo__/examples/arbitraryChildren.js
```diff
@@ -0,0 +1,28 @@
...
+  <img src="https://www.fillmurray.com/100/300" />
```
brentertz: Can this image be smaller?

packages/link/package.json
brentertz: Probably not.

brentertz:
Any thoughts to what happens if users add an `onClick` handler?  Does it get a `role="button"`, do we warn them, in dev, that they should use a Button instead?  Is it a documentation thing?

packages/link/src/Link.js
brentertz: I don't see a theme at all.  This component should have a theme of it's own.

packages/link/src/__demo__/examples/link.js
brentertz: This example also does not work as expected.  It goes to `http://0.0.0.0:8080/https://example.com`

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import React from 'react';
+import { Link as RouterLink } from 'react-router-dom';
```
brentertz: I don't think that we want to directly tie this package to react-router.  

One solution might be to have this component use a regular `a` tag, but allow the base component to be specified via a prop.  Then users could pass in a react-router `Link` or `NavLink` and that would get custom styled.  As long as all props, including the `to` prop,  are passed through to the base component, this should work. This would avoid any direct dependencies.  We would of course need some examples, demonstrating this functionality.

packages/link/src/Link.js
```diff
@@ -0,0 +1,66 @@
...
+import { createStyledComponent } from '@mineral-ui/component-utils';
+
+type Props = {
+  /** Content of the link. Can be anything */
+  children: MnrlReactNode,
+  /** URI tied to the Link */
+  to: string
```
brentertz: We should consider what other props to surface.  e.g. `href`, `title`, `target`, etc.

brentertz: react-router allows the `to` prop to be a string or an object.  My recommendation is likely to not include that prop here, but rather build a basic styled Link component that uses props based on standard HTML attributes.  If users choose to change the base component to a react-router Link, as described elsewhere, those props can be passed through via restProps, though they should not need declared.

kylegach:
Re: https://github.com/mineral-ui/mineral-ui/pull/211#issuecomment-319367427

I read a fascinating discussion about this decision the other day: https://github.com/alphagov/govuk_elements/pull/272

packages/site/src/components/__tests__/__snapshots__/App.spec.js.snap
coldpour: fixed

packages/link/src/Link.js
coldpour: fixed

packages/link/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```jsx
+import React from 'react';
+import Link from '@mineral-ui/link';
+
+export default function MyComponent() {
+  return (
+    <div>
+      <Link href="https://mineral-ui.com">
+        Best Site in the universe
```
kylegach: Not sure this is a good look.

packages/link/package.json
kylegach: Is `react-router-dom` still a peer dependency because of the demo example?

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+import React from 'react';
+import { createStyledComponent } from '@mineral-ui/component-utils';
+
+type Props = {
+  /** Content of the link. Can be anything */
```
kylegach: Please match the style of the other components (e.g. no period). I also don't think "Can be anything" is necessary, especially since you have an example demonstrating so.

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  children: MnrlReactNode,
+  /** Link renders an <a> by default but you can specify a different tag or component as the root (e.g. a React Router Link).*/
```
kylegach: We have a Default column; no need to also spell it out in the Description.

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  root?: MnrlReactNode
```
kylegach: How do you feel about `element`, instead? `root` doesn't seem very clear. Could just be me, though.

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+/**
+ * Link component
+ */
+export default function Link({ children, root = 'a', ...otherProps }: Props) {
```
kylegach: `otherProps` -> `restProps`, per the pattern used in other components.

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  if (otherProps.onClick) {
+    console.error(
```
kylegach: Will a `console.error` potentially stop a build in a consuming app? If so, is there _no_ legitimate reason to use an `onClick` on a link? Seems like there could be (tracking purposes come to mind).

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+      'You put an onClick on a Link. Maybe you want a Button instead?'
+    );
+  }
+  const Root = createStyledComponent(
+    root,
+    (props, theme) => {
```
kylegach: To follow the pattern in our other components, this function should be:

```js
(props, baseTheme) => {
  const theme = {
    Link_color: baseTheme.color_link,
    Link_color_active: baseTheme.color_link_active,
    Link_color_focus: baseTheme.color_link_focus,
    Link_color_hover: baseTheme.color_link_hover,
    Link_fontWeight: baseTheme.fontWeight_regular,
    Link_fontSize: `${2 * baseTheme.fontSize_base}px`

    ...baseTheme
  };

  return {
    color: theme.Link_color,
    fontWeight: theme.Link_fontWeight,
    fontSize: theme.Link_fontSize,

    '&:hover': {
      color: theme.Link_color_hover
    },
    '&:focus': {
      color: theme.Link_color_focus
    },
    // `:active` must be last, to follow LVHFA order:
    // https://developer.mozilla.org/en-US/docs/Web/CSS/:active
    '&:active': {
      color: theme.Link_color_active
    }
  };
}
```

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+      return {
+        color: theme.color_link,
+        fontWeight: theme.Link_fontWeight || theme.fontWeight_regular,
+        fontSize: theme.Link_fontSize || `${2 * theme.fontSize_base}px`,
+
+        '&:hover': {
+          color: theme.color_link_hover
+        },
+        '&:focus': {
```
kylegach: This focus style is not even remotely discernible. Please follow the `boxShadow` design used in Button (but with no padding).

packages/link/src/__demo__/examples/reactRouter.js
```diff
@@ -0,0 +1,45 @@
...
+  Link as ReactRouterLink,
+  Route,
+  Switch
+} from 'react-router-dom';
+
+export default {
+  title: 'Using React Router',
+  description: `In this example, we import { Link as ReactRouterLink } from 'react-router-dom' so that we can get client-side routing with our styles.`,
+  scope: { Link, ReactRouterLink, BrowserRouter, Switch, Route },
+  source: `<BrowserRouter>
```
kylegach: _Not relevant to this PR_: We thought of the need for examples with a hidden source, but didn't think of examples that need to hide the preview, like this one.

packages/link/src/__demo__/examples/states.js
```diff
@@ -13,27 +13,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import darken from 'polished/lib/color/darken';
```
kylegach: Should we be concerned that git apparently thinks this file is a renamed `packages/site/src/components/Link.js`?

packages/site/src/components/Footer.js
kylegach: ⬆️ Reminder

packages/site/src/components/__tests__/__snapshots__/App.spec.js.snap
```diff
@@ -912,6 +912,77 @@ exports[`App renders correctly 1`] = `
...
+          <Link {...simulate('focus')} href=\\"http://example.com\\">Focus</Link>
+          <Link {...simulate('active')} href=\\"http://example.com\\">Active</Link>
+        </DemoLayout>",
+                "title": "States",
+              },
+            ],
+            "slug": "link",
+            "title": "Link",
+          },
+          "world": Object {
```
kylegach: I think you need to update this snapshot after rebasing from master. `Hello`, `World`, and `HelloWorld` no longer exist.

riley:
@brentertz The solution we have right now is to have the component print a `console.error` with a warning (`console.warn` doesn't have a stack trace).

Another story we'll have to definitely add is a way to style the Link as a Button, since sometimes implementers are going to want a call to action "button" that takes the user to another page.

packages/link/src/Link.js
riley: maybe `rootElement`?

packages/link/src/Link.js
riley: It won't act like a thrown error if I'm understanding your question. We used `console.error` so the user could get a stack trace to get a hint where the component was implemented.

packages/link/README.md
riley: whoops

packages/link/package.json
coldpour: yes

coldpour: er sorry, that makes it a devDependency. it does not need to be installed by a consumer

packages/link/src/__demo__/examples/arbitraryChildren.js
```diff
@@ -0,0 +1,27 @@
...
+
+/* @flow */
+import Link from '../../Link';
+
+export default {
+  title: 'Arbitrary children',
+  description: 'A Link will render any children.',
+  scope: { Link },
+  source: `<Link href="http://example.com">
+  <img src="https://www.fillmurray.com/100/100" />
```
brentertz: Image should have an alt attribute

packages/link/src/__demo__/examples/states.js
```diff
@@ -13,27 +13,26 @@
...
-  forwardProps: ['to']
-});
-
-export default createStyledComponent('a', styles);
+export default {
+  title: 'States',
+  hideSource: true,
+  scope: { Link, DemoLayout, simulate },
+  source: `<DemoLayout>
+  <Link href="http://example.com">Don't Panic</Link>
```
brentertz: We should just point all of these links to http://mineral-ui.com

packages/link/src/Link.js
brentertz: Agreed.  Let's follow the pattern that exists in other components.

packages/link/src/Link.js
```diff
@@ -0,0 +1,65 @@
...
+  children: MnrlReactNode,
+  /** Link renders an <a> by default but you can specify a different tag or component as the root (e.g. a React Router Link).*/
+  root?: MnrlReactNode
+};
+
+/**
+ * Link component
+ */
+export default function Link({ children, root = 'a', ...restProps }: Props) {
+  if (restProps.onClick) {
```
brentertz: I like the intention of enforcing the distinction between buttons and link, but I think logging an error or even a warning can potentially disrupt developer workflows, and we would never want to do this in production.  I think that we'll need to solve this via documentation once we begin writing usage guidelines.

packages/link/src/__tests__/Link.spec.js
```diff
@@ -0,0 +1,37 @@
...
+import Link from '../Link';
+import examples from '../__demo__/examples';
+import testDemoExamples from '../../../../utils/test/testDemoExamples';
+
+function renderLink(props, children) {
+  return shallow(<Link {...props}>{children}</Link>);
+}
+
+describe('Link', () => {
+  let link;
```
brentertz: No need for this `let` . It can be a `const link` inside the `it` block.

packages/link/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "flow": "$npm_package_config_script_path/flow.sh",
+    "jest": "$npm_package_config_script_path/jest.sh",
+    "lint": "$npm_package_config_script_path/lint.sh",
+    "pack-ls": "$npm_package_config_script_path/pack-ls.sh",
+    "start": "DEMO_LIST_PATH=\"link\" $npm_package_config_script_path/start.sh",
+    "test": "$npm_package_config_script_path/test.sh"
+  },
+  "devDependencies": {
+    "@mineral-ui/component-utils": "^1.0.0",
+    "react-router-dom": "^4.1.2"
```
brentertz: This only needs to appear in the project's package.json

packages/link/src/__demo__/examples/link.js
```diff
@@ -0,0 +1,27 @@
...
+
+/* @flow */
+import Link from '../../Link';
+
+export default {
+  title: 'Basic link',
+  description: 'Links can be used in prose to create hypertext.',
+  scope: { Link },
+  source: `<p>
+Sensors indicate human life forms 30 meters below the planet surface. Stellar flares are increasing in magnitude and frequency. Set course for <Link target="_blank" href="http://memory-alpha.wikia.com/wiki/Rhomboid_Dronegar_sector_006">Rhomboid Dronegar 006</Link>, warp seven. There is no evidence of an advanced communication network. Total guidance system failure, with less than 24 hours reserve power. Shield effectiveness has been reduced 12 percent. We have covered the area in a spherical pattern which a ship without warp drive could cross in the given time.
```
brentertz: Please just reuse the lipsum that is used elsewhere on the site.  We want to try to keep the language standard/professional until we have defined a voice of our own.

brentertz:
@kylegach Can you please run the focus styles by the design team.  I personally think that they are a bit odd/unexpected.

brentertz:
I am still feeling a bit wishy-washy on the prop named, "element".  Can we please revisit it one last time?

brentertz:
Re: https://github.com/mineral-ui/mineral-ui/pull/211#issuecomment-320246081 — I updated it to simply use the browser’s focus styles, for now.

kylegach:
Re: https://github.com/mineral-ui/mineral-ui/pull/211#issuecomment-320246081 — I updated it to simply use the browser’s focus styles, until we receive a design spec.

PR #215, opened 2017-07-28
Icon doc

### LINKED ISSUE #214: [Icon] Testing
make sure icons work outside of the demo app

AC:

mineral icon can be successfully consumed by create-react app and instructions for how to do so are in the README

coldpour:
### Description

update icon README to describe how to use the mineral icon package

closes #214 

### Motivation and context

The old instructions had the wrong import path for `Icon`s. The import path was to the index, which is no longer the recommended way of importing. We have to recommend people use the direct file path because the way we export our components breaks webpack's flaky tree shaking ability. If people import the wrong way, they're going to get all the icons in their bundle whether they're using them or not.

### Types of changes

docs

### Checklist

### DIFF
--- a/packages/icon/README.md
+++ b/packages/icon/README.md
@@ -23,18 +23,20 @@ a generic Icon component to which you can provide your own SVG.
 
 ```jsx
 import React from 'react';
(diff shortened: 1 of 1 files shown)


coldpour:
closes #215

PR #228, opened 2017-08-07
[All] Adjust base font size

kylegach:
### Description
- Change `fontSize_base` from 8 to 16
- Fix a couple component style issues that resulted
- Update `borderRadius_1` theme value and remove unused `borderRadius_2-5` variables (can be pulled into a separate PR)
- Remove unused theme variables (can be pulled into a separate PR)

### Motivation and context

The `fontSize_base` of 8 was chosen for two reasons:

1. The design system is built on an 8px grid
1. It reduced the significant digits in our theme values and so reduced the risk of browser rounding differences

However, setting the root-level font-size to 8px for all components also has some issues. Most notably, this is what it looks like if you place a paragraph of text inside a Card (with no CardBlock):

Tiny card content

Using a `fontSize_base` of 16 instead would be much more expected by the consumer, as it is the default browser font-size. Additionally, each of the two reasons above can be addressed:

1. 16 is simply double the 8px grid value; the mental math isn’t too much harder
1. browser rounding differences can be mitigated by applying a strict (i.e. set heights & line-heights) layout strategy, which is likely necessary anyway (I confirmed this is the case for Button; no visual changes in either Edge or IE11).

### How to test

1. `cd mineral-ui/packages/site && npm start`
2. Confirm all components look as they should

### Types of changes

### Checklist

Algorithms all around me


packages/button/src/Button.js
```diff
@@ -147,6 +147,7 @@ const buttonStyles = (props, baseTheme) => {
     cursor: props.disabled ? 'default' : 'pointer',
     fontWeight: theme.Button_fontWeight,
     height: theme[`Button_size_${props.size}`],
+    lineHeight: 0.75, // theme.fontSize_base / parseFloat(theme.ButtonContent_fontSize_small)
```
kylegach: Without this line, the text of small and medium-sized Buttons is too low.

The meaningful lineHeight is applied to ButtonContent, so the root level lineHeight can be anything. I used Dev Tools to determine that anything below 0.87 fixes the issue. The code in the comment is my attempt to rationalize that magic number.

packages/link/src/Link.js
```diff
@@ -33,7 +33,6 @@ const linkStyles = (props, baseTheme) => {
     Link_color_hover: baseTheme.color_link_hover,
     Link_color_focus: baseTheme.color_link_focus,
     Link_color_active: baseTheme.color_link_active,
-    Link_fontSize: `${2 * baseTheme.fontSize_base}px`,
```
kylegach: At the moment, this is the only component that must be updated along with `fontSize_base`. :(

kylegach: Also, not sure if removing this line entirely is the way to go...

kylegach:
The Site will be messed up by this change. I didn't want to update that until after discussion though, because I'd like to get it using mineralTheme at the same time.

packages/button/src/Button.js
brentertz: We should probably include the actual calculation here and also do some experimentation/validation that it works across various values.

packages/link/src/Link.js
brentertz: I think we should keep this theme variable, but just set it to `baseTheme.fontSize_base`

coldpour:
I like the goal of this PR, to make the default font in arbitrary Card content more in line with what's expected. I'm unclear of the tradeoffs, but I trust your judgement @kylegach.

brentertz:
Some of the font sizes in the site are smaller than expected.
screen shot 2017-08-10 at 10 59 49 am

kylegach:
Yep: https://github.com/mineral-ui/mineral-ui/pull/228#issuecomment-320764144

brentertz:
Sorry, I missed your comment...

brentertz:
Close in favor of https://github.com/mineral-ui/mineral-ui/pull/248 ???

kylegach:
⬆️

PR #231, opened 2017-08-07
[Card] Pass all props to underlying root component

### LINKED ISSUE #229: [CardTitle] props not passed through
the CardTitle component won't let me put a className on it. I expect it to get passed through to the DOM so that I can attach my CSS to it, but it's getting removed.

kylegach:
### Description

I think the title covers it. This pattern should be used on all public components.

Closes #229 

### Motivation and context

See #229 

### How to test

1. Pass CardTitle a `className` (or any other non-specified, valid DOM) prop in a consuming app
1. Confirm that prop is rendered correctly in the DOM

### Types of changes

### Checklist

Quick

### DIFF
--- a/packages/card/src/CardTitle.js
+++ b/packages/card/src/CardTitle.js
@@ -91,10 +91,20 @@ const Title = createStyledComponent('h3', styles.title);
 /**
  * Card title component
  */
-export default function CardTitle({ children, meta, minor, subtitle }: Props) {
+export default function CardTitle({
+  children,
+  meta,
+  minor,
+  subtitle,
+  ...restProps
+}: Props) {
+  const rootProps = {
+    minor,
+    ...restProps
+  };
   const isMinor = minor || Boolean(meta);
   return (
-    <Root minor={minor}>
+    <Root {...rootProps}>
       {meta && <Meta>{meta}</Meta>}
       <Title minor={isMinor}>{children}</Title>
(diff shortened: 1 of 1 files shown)


packages/card/src/CardTitle.js
```diff
@@ -28,6 +28,8 @@ type Props = {|
   meta?: string,
   /** Title of the card */
   children: MnrlReactNode,
+  /** HTML class applied to the root element */
+  className?: string,
```
kylegach: I wasn’t sure if this was the right approach, or if we should simply pass along all props like we do for other components. I didn’t do that here, because CardTitle is only to be used inside of Card.

coldpour: this feels inconsistent since we don't explicitly have this on other components

kylegach: Yeah, I think I agree. Probably better to just pass along all props.

kylegach:
I pushed a second commit that simply passes along all props, rather than explicitly passing along only `className` (first commit).

brentertz:
The issue here is that any props passed to a public component should be passed to the underlying component root.  This is not limited to className.

packages/card/src/CardTitle.js
brentertz: All public components should pass props to the root.

brentertz:
Have you verified that the other Card components do the same?

PR #263, opened 2017-08-14
Add versioning strategy discussion

kylegach:
### Description
Documenting our discussion around versioning.

### Motivation and context

For posterity.

### Types of changes

Docs

### Checklist

Decide

### DIFF
--- a/docs/discussions/versioning.md
+++ b/docs/discussions/versioning.md
@@ -1,0 +1,164 @@
+# Versioning
+
+Our current versioning strategy seems to have us on a path of rapidly climbing version numbers for our components (as of this writing, Button is on 4.0.0). This is not ideal because:
+
+- A project with many major changes in a short period of time appears unstable to consumers.
+- Having our packages at these relatively disparate versions makes it difficult, if not impossible, to coordinate which versions work together.
+
+There's an argument to be made that we jumped the gun on going "1.0.0" too early. Since we still have so many breaking changes needing released, we aren't very stable, and the high version numbers imply otherwise.
+
+On the other hand, our current strategy makes it easy for our consumers to know what to expect, w/r/t minor vs. major changes.
+
+
+## Current Strategy
+
+We have followed [semver](http://semver.org/) as best as we could from day one. Additionally, we have considered a change to styles that affect the box-model (dimensions, margins, positioning, etc...) a major, breaking change.
+
+
+## Proposals
+
+### A. No Change
+
+Just keep on keeping on.
+
+
+### B. `0.x.x`
+
+Per a conversation with npm, we have three ways to do this:
+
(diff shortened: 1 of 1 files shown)


brentertz:
Whichever way we go, it will need documented for consumers somewhere more visibly than this document, e.g. in the readme and in the site.

kylegach:
Agreed 💯%.

brentertz:
Strategy B sounds like a 👑  PITA, particularly dealing with separating commits and the current release:manual vs. semantic process.

kylegach:
Agreed there, too. My suggestion is the third strategy: a new npm org.

docs/discussions/versioning.md
```diff
@@ -0,0 +1,164 @@
...
+   | ComponentUtils | 1.0.0 |
+
+   Then we release Table, which we don't believe to be stable yet...
+
+   | Component | Version |
+   | :-------- | :------ |
+   | Button | 1.0.0 |
+   | Card | 1.0.0 |
+   | ComponentUtils | 1.0.0 |
+   | Table | 0.0.0 |
```
brentertz: You can't actually release as `0.0.0`.  `0.1.0` is probably more accurate.

brentertz:
Your descriptions of how things play out down the road sound like a challenging maintenance burden.  I do not currently believe that our component versions need to be kept in sync and I believe that it would be quite difficult given our current architecture.

If you want all component versions to be in sync, then just change the lerna versioning mode to "Fixed/Locked".  Lerna's own docs however suggests "Independent mode ... makes sense for a group of components."  

What I am hearing is the desire to version the library, which is valid, though I think it is dangerous to attempt to predict a strategy for 2.0 based on design changes.  Following a 1.0 release, semver would need to win, regardless.  As our components are individually consumable, and we do not have an aggregate package, this is not entirely possible at present.  Perhaps we can solve it from a different angle.  Once we have an aggregate package, consumers that choose to use it won't ever have to deal with more than one version.  In the interim, peer dependencies within each package contain the associated supported versions.  Keeping it simple, consumers should always attempt to use the latest version of all mineral components.

If this is truly this important, we'll need to reconsider the strategy of publishing independent components and therefore also using lerna at all.  This wouldn't be the end of the world, but it would take some rework.

As the 1.0 ship has sailed, due to being honest about changes and using semantic versioning from day 1, I think that the only way to reset the baseline is to move to a new org.

brentertz:
@kylegach Do we need this PR anymore?  It has been resolved by switching to a single library package model rather than publishing independently packaged/versioned components.  This allows us to reset versioning at 0 and stay below 1.0.0 until we deem appropriate.  Documentation has been updated to include our new versioning strategy.

PR #265, opened 2017-08-14
[Site]: Document passthrough props behavior

### LINKED ISSUE #209: [Docs] pass-through props
as a consumer I want to be able to read in the documentation about how non-documented props are handled.

AC:
 a consumer can read the docs and understand how their props will behave and can successfully pass through their props

Other Considerations:
* add some info to props table
  * commonly used things like `className`
* what to do when no props table exists
* e.g.: "other properties, not documented, are applied to the root element"

brentertz:
### Description

Add note about undocumented props underneath the props table

The props section will now appear regardless if there are props specified for a component.  For packages that are not components, there is now a flag to hide the prop doc, called `hidePropDoc`.  Also tweaked `renderExamples` to not render as an array as it is less readable than just using a wrapper div and I could find no specific layout/css reasons to avoid it.

### Motivation and context

This is important information that consumers will want/need to know.

closes #209 

screen shot 2017-08-14 at 2 38 22 pm
screen shot 2017-08-14 at 2 38 33 pm

### How to test

Try it out in the demo.

### Types of changes

Minor update to site

### Checklist

### DIFF
--- a/packages/component-utils/src/__demo__/index.js
+++ b/packages/component-utils/src/__demo__/index.js
@@ -25,6 +25,7 @@ export default {
       'Sample implementations of various methods of styling and style overrides.'
   },
   examples,
+  hidePropDoc: true,
   slug: 'component-utils',
   title: 'ComponentUtils'
 };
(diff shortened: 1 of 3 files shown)


brentertz:
Doh, I just noticed that I need to hide the prop docs from package docs that are not components, e.g. `component-utils`.  Will update...

PR #286, opened 2017-08-17
Remove CLI deploy from release:semantic

kylegach:
### Description

With the switch to Netlify, we no longer need to manually deploy the site after release via the CLI. Instead, Netlify’s continuous deployment will automatically deploy when the chore(publish) commit is merged into master.

### Motivation and context

It’s wasteful to deploy the same thing twice. Also, authentication can be an issue for the CLI deploy.

### How to test

1. `cd mineral-ui && npm run release:semantic`
1. Step through options
1. Confirm that `netlify deploy` never runs
1. Confirm that the site still deployed via continuous deployment

### Types of changes

Build-related

### Checklist

Facepalm

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "lint": "npm run eslint",
     "pack-ls": "lerna run pack-ls",
     "release:manual": "npm run build && npm run test && lerna publish --message='chore(publish): Publish'",
-    "release:semantic": "npm run release:manual -- --conventional-commits && npm run deploy:site:dist",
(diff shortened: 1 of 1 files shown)


riley:
The `release:semantic` script just errors out if there's nothing to publish. The only way to really test this is to add it to an actual publish, but that seems like mixing concerns. I did a text search on the code base and there's nowhere else that's calling `netlify deploy` or `deploy:site`.

PR #275, opened 2017-08-16
[Site] Placeholder pages

riley:
### Description

- Add GettingStarted, Guidelines, and Theming pages
- Add links to Nav
- Add Link component in Site (to abstract away React Router)
- Extract Table from PropTable in Site

### Motivation and context

Closes #219 

### Screenshots or videos, if appropriate

screen shot 2017-08-28 at 4 20 06 pm

### How to test

1. `cd mineral-ui/packages/site && npm start`
2. Confirm the new nav links work

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support)
* [x] Automated tests written and passing
* [x] [Accessibility](http://webaim.org/intro) and [inclusivity](https://24ways.org/2016/what-the-heck-is-inclusive-design/) considered
* [x] Rendering performance (initial load time & 60fps) and [perceived performance](http://blog.teamtreehouse.com/perceived-performance) considered [n/a]
* [x] Documentation created or updated
* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change [n/a]

### How does this PR make you feel?

Blank

Issue #269, opened 2017-08-15
[Site] Create Getting Started page


kylegach:
As a consumer of or contributor to Mineral UI, I want a page that explains how to get started, so that I can do so as quickly and smoothly as possible.

AC:
1. Create page
1. Page should contain
    1. Getting started from consumers perspective
    1. Getting started from contributors perspective

PR #290, opened 2017-08-18
[Site] Getting Started

kylegach:
connects #269 

### Description
- Add real content to Getting Started page
- Add CodeBlock component

### Motivation and context

People need to know how to get started, even if they don’t visit out GitHub repo.

### How to test

1. `cd mineral-ui/packages/site && npm start`
2. It should redirect you to /getting-started (but that's an issue with #275 if it doesn’t work), if not, navigate there
3. Read the content

### Types of changes

### Checklist

Get started


packages/site/src/components/CodeBlock.js
```diff
@@ -0,0 +1,47 @@
...
+  language?: string
+|};
+
+const Pre = createStyledComponent('pre', ({ theme }) => ({
+  fontSize: theme.fontSize_ui,
+  maxHeight: `${parseFloat(theme.spacing_quad) * 10}em`,
+  overflow: 'auto'
+}));
+
+export default function CodeBlock({ children, language }: Props) {
```
brentertz: You cannot get syntax highlighting via `prism` in this manner.  You actually need to transform the content via `prism`.

packages/site/src/pages/GettingStarted.js
```diff
@@ -0,0 +1,149 @@
...
+import PageLayout from '../components/PageLayout';
+
+const Callout = createStyledComponent('p', ({ theme }) => ({
+  backgroundColor: theme.color_gray_10,
+  borderLeft: `4px solid ${theme.borderColor}`,
+  padding: `${theme.spacing_single} ${theme.spacing_double}`
+}));
+
+export default function GettingStarted() {
+  return (
```
brentertz: The content of this getting started page is going to change substantially in the next 2 weeks due to the architecture updates.  We should probably hold off on including this content, in order to avoid the need to simply update it again.

packages/site/src/components/CodeBlock.js
kylegach: Addressed.

packages/site/src/pages/GettingStarted.js
kylegach: It’ll just be content changes, which shouldn’t take long to update and neither of us mind doing so. Additionally, we’d rather not merge this website work (on `web-mvp` branch) without a Getting Started page, as it would just be Guidelines and Theming without it.

Issue #271, opened 2017-08-15
[Site] Build out Component page IA


kylegach:
As a consumer of Mineral UI, I want a more complete component page, so that I have access to all of the information I need at a component level.

AC:
1. Build out the slots for the content and IA represented [here](https://github.com/mineral-ui/mineral-ui/issues/41)

PR #300, opened 2017-08-24
[ComponentUtils] Reorganize demo to provide api examples

riley:
### Description
We wanted to update the `component-utils` docs pages to expose the API more directly while maintaining test coverage.

- add `id` to all demo examples
- make section headers linkable
- update propTypes of ComponentDoc and ComponentDocExample
- update testDemoExamples to use `id`
- update snapshots
- update unit tests

### Motivation and context
We wanted to have good content to link to while building out the Theming guide pages.

### How to test

1. `cd mineral-ui/packages/component-utils && npm start`
2. Make sure all pertinent API information is available
3. Ensure you can go directly to the anchor link for each section

### Types of changes

- Docs

### Checklist

Alt text here


packages/component-utils/src/__demo__/components/Code.js
```diff
@@ -15,16 +15,12 @@
...
-    <ThemeProvider theme={{ color_primary: 'mediumvioletred' }}>
-      <Sample />
-    </ThemeProvider>`
-};
+export default createStyledComponent('code', ({ theme }) => ({
+  backgroundColor: theme.color_gray_20,
+  borderRadius: theme.borderRadius_1,
+  fontFamily: theme.fontFamily_monospace,
+  padding: theme.spacing_half,
+  whiteSpace: 'nowrap'
```
brentertz: The `whiteSpace: 'nowrap'` on the `code` component causes the contained text to extend off to the right of smaller screens.  Is it necessary?

brentertz: Relatedly, I noticed that it is the `flexbox` styles on the `header` element in the `ComponentDoc` that is causing the extra scroll width on small screens.

packages/site/src/components/ComponentDocExample.js
```diff
@@ -80,13 +81,14 @@ export default class ComponentDocExample extends Component {
...
       hideSource,
+      id,
       scope,
       source,
       title
     } = this.props;
 
     return (
-      <Root className={className}>
+      <Root className={className} id={id}>
```
brentertz: In addition to making these elements targetable, a common idiom is to include links as well.  This way, one can simply copy the link url and use it elsewhere - e.g. share it, paste it in docs, bookmark it, etc.  See the headers in https://facebook.github.io/react/docs/react-api.html as an example.

packages/component-utils/src/__demo__/examples/createStyledComponent.js
```diff
@@ -15,13 +15,44 @@
...
+        <Link href="https://github.com/threepointone/glamor/blob/master/docs/howto.md">
+          object of style rules
+        </Link>{' '}
+        or a function that accepts props and context and returns an object of
+        style rules
+      </p>
+      <p>
+        <Code>options</Code> — An object of optional configuration. A mix of
+        Mineral UI and{' '}
+        <Link href="https://github.com/paypal/glamorous/#glamorous-api">
```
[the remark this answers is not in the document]
kylegach: Good catch!

packages/component-utils/src/__demo__/examples/createStyledComponent.js
```diff
@@ -15,13 +15,44 @@
...
+          Glamorous options
+        </Link>. Most common uses are setting a display name on your component,{' '}
```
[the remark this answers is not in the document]
kylegach: Hmm... I see where you’re coming from, but I disagree. We added that text specifically to call out those options (particularly “forwardProps”), so that if someone went looking for “props”, they’d find that easily.

packages/component-utils/src/__demo__/components/Code.js
kylegach: Good catch; thanks. That issue is worse than fixing the wrapping within the prose, so I removed it.

The `header` issue will be addressed with #271.

packages/site/src/components/ComponentDocExample.js
kylegach: We opted to keep it simple for this first pass, but I was able to add this feature with about 10 mins work, so in it goes. 😃

PR #319, opened 2017-09-06
Update changelog generation

brentertz:
### Description

Update changelog generation
* Only include things from `mineral-ui` package, not previous `@mineral-ui` packages
* Remove changelog header, as it is more trouble than it is worth
* Don’t append entries. While that kept the header, it sorted commits in reverse

### Motivation and context

I created the changelog following recent commits and it still included all previous commits from the history of the repo, which is not what we want.  We only want to show entries since the creation of the `mineral-ui` package.

### How to test

1. `npm run release`
2. When prompted to publish, select **no**
3. View changelog - it should have new entries only
4. Revert local changes caused by release script.  "a git tag and 2 commits"

### Types of changes

build - update release script for changelog generation

### Checklist

* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change - [n/a]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,3 @@
-# Change Log
-
-All notable changes to this project will be documented in this file.
 <a name="0.1.0-alpha.0"></a>
 # 0.1.0-alpha.0 (2017-09-05)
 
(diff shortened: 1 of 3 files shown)


riley:
excellent! works for me. 👍

PR #323, opened 2017-09-08
[website]: Use Glamor to inject global css

brentertz:
### Description

* Uses `Glamor` to inject global CSS rather than `css-loader` and `style-loader`
* Improvement of "live" code block legibility on small screens
    * Removes forced line wrapping inside code blocks as this no longer appears to be needed

### Motivation and context

We already use Glamor, so there is no need for those other dependencies and extra configuration.

@kylegach - This removes ~112 packages from node_modules including all of those `postcss`modules that you were inquiring about.

### How to test

1. View contents of styles in header and note that the global styles are injected
1. View "live" code blocks in smaller screens and ensure that they don't break layout, content is horizontally scrollable, lines don't wrap, and that they are generally more legible

### Types of changes

Refactor of global css loading technique.
Improvement of "live" code block legibility on small screens

### Checklist


src/website/index.css
```diff
@@ -1,7 +0,0 @@
-body {
-    margin: 0;
-}
-pre.prism-code {
```
kylegach: I think these styles are only no longer needed when there is no sidebar, otherwise it squeezes the sidebar to be too narrow.

Rather than making them global, though, maybe we should make a `Pre` component in the site, and use that in `ComponentDocExample` and `Markdown` (those are the only places using codeblocks, I think).

src/website/index.js
```diff
@@ -16,15 +16,19 @@
...
 import ThemeProvider from '../ThemeProvider';
 import App from './app/App';
 import demos from './app/demos';
-require('./index.css');
-require('react-live/react-live.css');
+import reactLiveCss from 'raw-loader!react-live/react-live.css';
+
+// Global CSS
+css.global('body', { margin: 0 });
+css.insert(reactLiveCss);
```
kylegach: These don’t seem to be inserted in the deploy preview? They are when I run it locally, though.

brentertz: What, why?

src/website/index.css
brentertz: Good suggestion to create a new component to take it a step further.  I may try it if it does not blow up the scope.

brentertz: I still think that the word-wrap/line-break styles should be removed.  It really helps legibility.  

The issue with the sidebar likely needs resolved differently as the current styles also cause problems with the Popover/boundary demo which uses the ScrollBox component.  It currently appears to be due the main layout's flexbox configuration.  Perhaps there is a fix where we can continue to use flexbox; alternatively we may need to fall back to floats.

I will happily demo this for you to make things more clear.

brentertz:
Closing...

Style changes have been extracted and committed to master via 1628da74b3fcb467182c70b1878e8f6dbb8eef2e

It turns out that Glamor's `css.insert` only works with a single rule, which may have had something to do with why it worked locally and not on netlify.

PR #334, opened 2017-09-13
[utils] createThemedComponent only uses ThemeProvider to pass custom theme

kylegach:
### Description

In `createThemedComponent`, we’re passing along the custom theme to be applied to both the `ThemeProvider` used by the helper _and_ the wrapped component. Per [glamorous docs](https://glamorous.rocks/advanced#theming) (the end of the code block), the surrounding theme is ignored when you pass a theme directly to a component via props. So, this change is to _not_ pass the custom theme directly to the wrapped component.

This PR also:
- Updates the Sample component to use proper theme variables
- Updates the `utils` examples to be more clear and helpful
- Updates snapshots

### Motivation and context

`createThemedComponent` currently does not work on any Mineral UI components. This fixes that.

### How to test

1. Navigate to utils
1. Confirm that the `createThemedComponent` example is working correctly
1. Open that example in your editor
    1. Change Sample to Button (remember to import and add to `scope`)
    1. Confirm that the example still works as expected
    1. (Note: all of the above is tested via snapshots)

### Types of changes

### Checklist

Repair man man man


brentertz:
Interesting, and somewhat sad that it existed in the first place.  So did your determine why this was working with our `Sample` themed component and not on others?

I am not entirely sure how to add automated test coverage for this, outside of our snapshot testing.  Off the top of my head, I am not aware of a way, with glamorous, to read the theme from the rendered component to be used within a test.  Ideally, we could read it and just assert that it has the expected contents.

FYI, I updated the PR to avoid mutating `restProps`.  I originally did it by destructuring assignment and ignoring the extra "theme" variable, but then thought that we should really be better about this across the entire project, and doing it sooner rather than later is usually a good idea, so thus introduced [immutability-helper](https://github.com/kolodny/immutability-helper), formerly known as [react-addons-update](https://facebook.github.io/react/docs/update.html).

PR #328, opened 2017-09-11
[Website] typography page

riley:
connects #240

### Description
Create a Typography page with demos and a list of related tokens.

### Motivation and context
We'd like to document the design system thinking with regard to Typography. This guideline page attempts to show designers what the thought process is in choosing our fonts, and how to choose different styles for different intentions when designing a UI. For (theme) developers, we provide some style hook documentation.

### How to test
1. `cd mineral-ui && npm start`
2. Navigate to Typography in the side menu
3. Look over the content. Are there spelling errors or inconsistencies?

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support)
* [x] Automated tests written and passing
* [x] [Accessibility](http://webaim.org/intro) and [inclusivity](https://24ways.org/2016/what-the-heck-is-inclusive-design/) considered
* [x] Rendering performance (initial load time & 60fps) and [perceived performance](http://blog.teamtreehouse.com/perceived-performance) considered
* [x] Documentation created or updated
* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change [n/a]

### How does this PR make you feel?

Alt text here

PR #354, opened 2017-09-22
[Website] Display markdown typography according to spec

riley:
### Description
Added proper shades of gray to Heading component and create a Paragraph element per the design spec.

Also add Cyrillic and Greek glyph support to Open Sans, since the site needs it in a couple of places.

### Motivation and context
We need visual consistency when comparing our Guidelines pages, and only #328 had the typography updates. This PR pulls those changes out so that there are fewer facets to think about when discussing hierarchy on the content pages.

### How to test

2. Navigate to Guidelines > Theming
3. see that the h1 is "Mineral black" instead of `#000`, and the h2 is gray.
4. Markdown-generated body copy is now the `prose` font style.

### Types of changes

### Checklist

Alt text here

### DIFF
--- a/src/website/app/Heading.js
+++ b/src/website/app/Heading.js
@@ -31,6 +31,12 @@ type Props = {
 };
 
 const componentTheme = baseTheme => ({
+  Heading_color_1: baseTheme.color_text,
+  Heading_color_2: baseTheme.color_gray_80,
+  Heading_color_3: baseTheme.color_text,
+  Heading_color_4: baseTheme.color_text,
+  Heading_color_5: baseTheme.color_text,
+  Heading_color_6: baseTheme.color_gray_80,
   Heading_fontSize_1: baseTheme.fontSize_h1,
   Heading_fontSize_2: baseTheme.fontSize_h2,
   Heading_fontSize_3: baseTheme.fontSize_h3,
@@ -57,6 +63,7 @@ const headingStyles = ({ level, theme: baseTheme }) => {
   let theme = componentTheme(baseTheme);
 
   return {
+    color: theme[`Heading_color_${level}`],
     fontSize: theme[`Heading_fontSize_${level}`],
     fontWeight: theme[`Heading_fontWeight_${level}`],
     margin: `${theme[`Heading_marginMultiplier_${level}`] *
--- a/src/website/app/Paragraph.js
+++ b/src/website/app/Paragraph.js
@@ -1,0 +1,63 @@
+/**
+ * Copyright 2017 CA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


src/website/app/Paragraph.js
```diff
@@ -0,0 +1,60 @@
...
+    fontSize: theme[`Paragraph_fontSize_${type}`],
+    lineHeight:
+      type === 'prose'
+        ? theme.Paragraph_lineHeight_prose
+        : theme.Paragraph_lineHeight_normal
+    // margins?
+  };
+});
+
+export default function Paragraph({ children, ...rest }: Props) {
```
brentertz: Our convention in this case is to call the rest param, `...restProps`

src/website/app/Paragraph.js
```diff
@@ -0,0 +1,60 @@
...
+  let theme = componentTheme(baseTheme);
+
+  return {
+    color: theme.Paragraph_color,
+    fontSize: theme[`Paragraph_fontSize_${type}`],
+    lineHeight:
+      type === 'prose'
+        ? theme.Paragraph_lineHeight_prose
+        : theme.Paragraph_lineHeight_normal
+    // margins?
```
brentertz: Comment needed?

src/website/app/Paragraph.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+/* @flow */
+import React from 'react';
+import { createStyledComponent } from '../../utils';
+
+type Props = {
+  /* rendered chilren */
+  children: MnrlReactNode,
+  /* lorem ipsum */
```
brentertz: Actual description for `type`

src/website/app/Paragraph.js
```diff
@@ -0,0 +1,60 @@
...
+  type: 'ui' | 'prose' | 'mouse'
```
brentertz: Our convention is to use the word, `variant` rather than `type`.

src/website/app/Heading.js
```diff
@@ -31,6 +31,12 @@ type Props = {
 };
 
 const componentTheme = baseTheme => ({
+  Heading_color_1: baseTheme.color_text,
+  Heading_color_2: baseTheme.color_gray_80,
+  Heading_color_3: baseTheme.color_gray_80,
+  Heading_color_4: baseTheme.color_gray_80,
+  Heading_color_5: baseTheme.color_text,
```
brentertz: This is an odd convention, no? (The `h5` color in particular...) I see that this matches the spec, but can we validate that it is not a mistake?

src/website/app/Paragraph.js
brentertz: Do we need a default value for `type/variant`?

PR #368, opened 2017-09-29
[Site] Force new routed navigations to scroll to top

kylegach:
### Description

Wrap the app in a `ScrollToTop` component (code from [react-router's site](https://reacttraining.com/react-router/web/guides/scroll-restoration)) that does exactly what it says on the tin.

### Motivation and context

Navigating to a new page in our site would, frustratingly, remember your scroll position rather than scrolling to the top automatically, as everyone expects. This fixes that.

### Demo

http://router-scrolling--mineral-ui.netlify.com/getting-started

### How to test

1. Open the site
2. Scroll down a bit and note what is visible
3. Navigate to another page
4. Confirm that the new page is automatically scrolled to the top
5. Hit the back button
6. Confirm that your scroll position was remembered for the previous page

### Types of changes

Docs

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support) **See note**

In Edge & IE 11, the scroll position is not remembered when going back; it auto-scrolls to the top. This is a regression from normal behavior (though arguably better than our site's current behavior), based on some testing on outside websites.

Pleased or satisfied

### DIFF
--- a/src/website/app/App.js
+++ b/src/website/app/App.js
@@ -15,7 +15,9 @@
  */
 
 /* @flow */
-import React from 'react';
+import React, { Component } from 'react';
+import { withRouter } from 'react-router';
+import { canUseDOM } from 'exenv';
 import { createStyledComponent } from '../../utils';
 import ComponentDoc from './pages/ComponentDoc';
 import createKeyMap from './utils/createKeyMap';
@@ -24,8 +26,10 @@ import _Nav from './Nav';
 import Router from './Router';
 
 type Props = {|
+  children?: any,
   className?: string,
-  demos: Object | Array<Object>
+  demos: Object | Array<Object>,
+  location?: any
 |};
 
 const styles = {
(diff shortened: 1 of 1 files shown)


src/website/app/ScrollToTop.js
```diff
@@ -0,0 +1,41 @@
...
+  location?: any
+};
+
+// https://reacttraining.com/react-router/web/guides/scroll-restoration
+class ScrollToTop extends Component {
+  props: Props;
+
+  componentDidUpdate(prevProps) {
+    if (this.props.location !== prevProps.location) {
+      window.scrollTo(0, 0);
```
brentertz: This will prolly break SSR, once we start pre-rendering, unless it is wrapped in a type check.

src/website/app/ScrollToTop.js
```diff
@@ -0,0 +1,41 @@
...
+/* @flow */
+import { Component } from 'react';
+import { withRouter } from 'react-router';
+
+type Props = {
+  children?: any,
+  location?: any
+};
+
+// https://reacttraining.com/react-router/web/guides/scroll-restoration
```
brentertz: Unnecessary comment/reference.  The code speaks for itself.

src/website/index.js
```diff
@@ -31,9 +32,11 @@ simulations(true);
 
 render(
   <BrowserRouter>
-    <ThemeProvider>
-      <App demos={demos} />
-    </ThemeProvider>
+    <ScrollToTop>
```
brentertz: Can this go inside of ThemeProvider rather than around it or perhaps just be included inside the App rather than as a separate component?

src/website/app/ScrollToTop.js
kylegach: Removed.

src/website/app/ScrollToTop.js
kylegach: Good catch, thanks. Added a check.

src/website/index.js
kylegach: Good call. I changed App to a class, added the relevant code to its `componentDidUpdate`, and exported `withRouter(App)`.

riley:
The only thing I would add if possible is that on the Theming page, there is a link to `createStyledComponent`. When you click this link, the page navigates, but does not scroll down to that section. However, If you copy the link and paste it into a new window, it works. Not sure if this PR includes that case.

kylegach:
@riley — I believe that to be a separate issue. I’m not sure why anchor links on that page don’t seem to work, but other anchor links have been tested and do work with this PR.

PR #371, opened 2017-09-29
[Site] Add routed page for each demo example

kylegach:
### Description

* Add routed page for each demo example
* Make each example title link to its page
* Add a chromeless mode, triggered by adding "?chromeless" to the end of the URL
* Refactor use of react-live components into a standalone LiveProvider component
* Change ComponentDocExample to a stateless functional component from a class
* Remove un-used Anchor styles in ComponentDocExample and other style improvements

### Motivation and context

When developing, it can be a pain to have multiple instances of the component on the screen. You could hack around this by editing the examples index, but this is _way_ nicer. The chromeless mode can be used later for visual diff testing.

http://example-permalink--mineral-ui.netlify.com/components/button

Note the URLs here...

screen shot 2017-09-29 at 4 22 10 pm

screen shot 2017-09-29 at 4 22 34 pm

### How to test

1. Open the site and navigate to a [component page](http://example-permalink--mineral-ui.netlify.com/components/button)
2. Click an example title to navigate to that example’s page
3. Add "?chromeless" to the end of the URL to switch to chromeless mode

### Types of changes

Docs

### Checklist

Ahhhhhh

### DIFF
--- a/src/website/app/App.js
+++ b/src/website/app/App.js
@@ -17,6 +17,7 @@
 /* @flow */
 import React, { Component } from 'react';
 import { withRouter } from 'react-router';
+import { Redirect, Route, Switch } from 'react-router-dom';
 import { canUseDOM } from 'exenv';
 import { createStyledComponent } from '../../utils';
 import ComponentDoc from './pages/ComponentDoc';
@@ -90,13 +91,28 @@ class App extends Component {
 
     return (
       <div>
-        <Root className={className}>
-          <Nav demos={siteDemos} />
-          <Main>
-            <Router demos={siteDemos} />
-            <Footer />
-          </Main>
-        </Root>
+        <Switch>
+          <Route
+            exact
+            strict
+            path="/:url*"
+            render={props => <Redirect to={`${props.location.pathname}/`} />}
+          />
+          <Route
+            render={route => {
+              const isChromeless = route.location.search === '?chromeless';
+              return isChromeless
+                ? <Router demos={siteDemos} />
+                : <Root className={className}>
+                    <Nav demos={siteDemos} />
+                    <Main>
+                      <Router demos={siteDemos} />
+                      <Footer />
+                    </Main>
+                  </Root>;
+            }}
+          />
+        </Switch>
       </div>
     );
   }
--- a/src/website/app/ComponentDocExample.js
+++ b/src/website/app/ComponentDocExample.js
@@ -15,12 +15,12 @@
  */
 
 /* @flow */
-import React, { Component } from 'react';
-import dedent from 'dedent';
-import { LiveProvider, LiveEditor, LiveError, LivePreview } from 'react-live';
-import { createStyledComponent } from '../../utils';
+import React from 'react';
+import { createStyledComponent, getNormalizedValue } from '../../utils';
 import Callout from './Callout';
 import Heading from './Heading';
+import Link from './Link';
+import LiveProvider from './LiveProvider';
 import Markdown from './Markdown';
 
 type Props = {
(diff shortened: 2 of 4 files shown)


src/website/app/Router.js
```diff
@@ -54,6 +57,25 @@ export default function Router({ demos }: Props) {
     <Switch>
       {routes}
       <Route
+        path="/components/:componentId/:exampleId"
+        render={route => {
+          const componentId = route.match.params.componentId;
```
brentertz: You could destructure both componentId and exampleId here and then reference them thereafter.

src/website/app/ComponentDocExample.js
```diff
@@ -90,7 +95,9 @@ export default class ComponentDocExample extends Component {
     return (
       <Root className={className} id={id}>
         <Title level={3} id={id}>
-          {title}
+          <TitleLink to={`./${id}`}>
```
brentertz: I really like how the title links to the standalone example, while the # links to the target without leaving the page. 👍

src/website/app/App.js
```diff
@@ -73,13 +74,20 @@ export default function App({ className, demos }: Props) {
...
-        <Main>
-          <Router demos={siteDemos} />
-          <Footer />
-        </Main>
-      </Root>
+      <Route
+        render={route => {
+          const isChromeless = route.location.search === '?chromeless';
+          return isChromeless
+            ? <Router demos={siteDemos} />
```
brentertz: Would it be possible to add a wrapper with some padding when in chromeless mode?  It is otherwise unbearable to use with stuff butting right up agains the viewport.

src/website/app/App.js
```diff
@@ -73,13 +74,20 @@ export default function App({ className, demos }: Props) {
...
-      <Root className={className}>
-        <Nav demos={siteDemos} />
-        <Main>
-          <Router demos={siteDemos} />
-          <Footer />
-        </Main>
-      </Root>
+      <Route
+        render={route => {
+          const isChromeless = route.location.search === '?chromeless';
```
brentertz: Yay, just like Sombrero!

src/website/app/LiveProvider.js
```diff
@@ -0,0 +1,60 @@
...
+  LiveEditor,
+  LiveError,
+  LivePreview
+} from 'react-live';
+import { createStyledComponent } from '../../utils';
+
+type Props = {
+  backgroundColor?: string,
+  hideSource?: boolean,
+  isChromeless?: boolean,
```
brentertz: Maybe this should be "chromeless" rather than "isChromeless"?  We seem to have been reserving the "isNNN" convention for state related booleans and using raw "NNN" for props related booleans.

src/website/app/Router.js
kylegach: 👍

src/website/app/ComponentDocExample.js
kylegach: Glad you like that pattern! I was debating a few different tacts.

src/website/app/LiveProvider.js
kylegach: Addressed.

src/website/app/App.js
kylegach: Yes, it would be trivial to add such a wrapper. However, that wrapper would then apply to the individual example chromeless pages too, without some _non-trivial_ routing/rendering logic. Those example pages already have [a small amount of padding added to their LivePreview](https://github.com/mineral-ui/mineral-ui/pull/371/files#diff-85abe98e739c3b9285bf8adcb93201ceR20); adding yet more padding would result in too much.

Chromeless mode is only intended for the individual example case. It happens to work on other pages because of its implementation details, but there is little-to-no value in optimizing those cases, imo.

PR #384, opened 2017-10-03
chore(flow): Flow improvements

### LINKED ISSUE #78: [Flow] Custom Types
As a consumer using Flow, I want to understandable Flow types in Mineral. I would like to not have to define custom types.

AC:
* MnrlReactNode and other custom flow types should show up as standard react flow types
* Consumers do not get flow errors for our custom types.

Other Considerations:
* Button icon props use react $element should they use Mineral types
* check in with Brent before pulling.
* We may want to publish our types to flow-typed
* Likely to remove custom types and duplicate/replace definitions

brentertz:
### Description
Flow improvements

* Get rid of custom “mnrl” flow-types - Yay!
* Update flow-bin which requires syntax changes
* Update babel-plugin-flow-react-proptypes
* Re-create flow-typed stubs
* Fix a bazillion flow issues. A couple confusing $FlowFixMe’s remain
* Still not publishing flow-types

### Motivation and context

Our flow types were confusing and getting a bit out of hand.

closes #78 

### How to test

* `npm i` & `npm test` and see if you get any flow errors
* Review changes and ensure that types make sense

### Types of changes

Mostly flow updates, but also includes some minor code and documentation updates too.

### Checklist

* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change - [n/a] . **Will do this once we decide to publish the flow-types**


src/website/app/pages/index.js
```diff
@@ -32,23 +32,23 @@ type Pages = Array<Page>;
 
 const pages: Pages = [
   {
-    title: 'Getting Started',
     component: GettingStarted,
+    title: 'Getting Started',
     path: '/getting-started'
```
kylegach: It looks like you moved `title` down to alphabetize, but now it goes c, t, p?

src/website/app/App.js
```diff
@@ -26,12 +26,12 @@ import Footer from './Footer';
 import _Nav from './Nav';
 import Router from './Router';
 
-type Props = {|
```
kylegach: Curious why you removed the strict bars from all of the website components.

src/utils/createStyledComponent.js
```diff
@@ -19,7 +19,10 @@ import glamorous from 'glamorous';
 import componentStyleReset from './componentStyleReset';
 
 export default function createStyledComponent(
-  element: MnrlReactNode,
+  element:
+    | React$StatelessFunctionalComponent<*>
+    | Class<React$ComponentType<*>>
+    | string,
```
kylegach: That looks fun to figure out. 😳

src/website/app/App.js
brentertz: They cause flow errors in some places when we are passing prop objects to components when those objects themselves are not strictly typed to match the props of the receiving component.

Also, the root prop spreading pattern that we use in the mineral components is useful and I don't see any reason why not to follow that pattern in the website.  Do you?

src/website/app/pages/index.js
brentertz: yep, good catch

src/website/app/App.js
kylegach: No, I agree with your choice. :+1: I was just curious why. It's nice to have the same pattern everywhere, too.

PR #307, opened 2017-08-31
[All] Markdown support, update component page

riley:
connects #271 

### Description
Markdown support for many places throughout the Site. Also some component page organization and general improvements to the Site.

- Add Markdown component using marksy for markdown support
- Refactor component data to use markdown and remove unused data
- Refactor non-component pages to use markdown
- Hide 'states' examples from prod
- Expose component theme variables
- Extract VariableTable from Theming page
- Add When/How To Use & Best Practices section
- Add fake content for more realistic layout
- Add Heading component and replace existing headings with it
- Add Section & Callout components
- Consistent padding around pages
- Move ComponentDoc to pages directory
- Fix snapshot exclusions that were using title instead of id (explains the huge number of deleted lines)

### Motivation and context
We needed to think through the visual hierarchy of the Component page, and add some ease-of-use things while we were in there.

### How to test

1. `cd mineral-ui/packages/site && npm start`
2. View any of the Component docs pages
3. Markdown works in many of the prose sections

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support) (see below)
* [x] Automated tests written and passing
* [x] [Accessibility](http://webaim.org/intro) and [inclusivity](https://24ways.org/2016/what-the-heck-is-inclusive-design/) considered
* [x] Rendering performance (initial load time & 60fps) and [perceived performance](http://blog.teamtreehouse.com/perceived-performance) considered (see below)
* [x] Documentation created or updated
* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change [n/a]

Re: browser support — We’re using CSS Grid for the columns of Best Practices. This does not work in IE11 or Edge. The former will never work; the latter will work [at some point in the future](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/gridupdate/?q=grid%20update).

Re: rendering performance — The pages seem to load a little slower with these changes. Guessing the culprit is the many Markdown content areas. Since this is “just” the site and not a public component, it seemed worth the minor slowdown. Maybe a future story could investigate a fix.

### How does this PR make you feel?

Writing

PR #399, opened 2017-10-05
[Theme] [Website] [All] Rework spacing variables

### LINKED ISSUE #340: [Theme] Additional spacing tokens
### Context

Provide easy-to-implement spacing tokens that solve common use-cases. Styleguide is suggesting primary spacing on 8p `0.5rem`?, with shims at 2p  and 4p. We redid `icon` to remove internal padding so everything lines up. We'll file issues where this affects existing components.

https://ca.invisionapp.com/share/Z4D71KSPX

AC:
- themeVariables match the spec

kylegach:
### Description

> Note for reviewers: You'll want to review this commit-by-commit, for the reasons detailed below.

#### Commit 1
- Add `space_inline_xx`, `space_inset_xx`, and `space_stack_xx` theme variables, per [spec](https://ca.invisionapp.com/d/main#/console/11070131/250301138/preview)
- Remove `spacing_xx` theme variables
- Update components to use the new variables
- Remove CardRow and cardTheme, which should have been done with #307 
- Add an exported `componentTheme` to MenuGroup and reference appropriately

#### Commit 2
- Update website components, including anything in `/demos/`, to use the new variables
- Refactor LiveEditor maxHeight
- Various small style improvements

#### Commit 3
- Update snapshots after modifying theme
    - Every component snapshot should have changed because the theme changed, but the _styles_ should not change at all
        - The kinda-exception to this is the Custom Render example, which has some styles in the source that reference the theme

#### Commit 4
- While updating and reviewing snapshots, I realized that the Menu spec was testing the snapshots for not just Menu, but all of the Menu* components. This corrects that. It's a separate commit because it makes the snapshots really hard to review for meaningful changes in the previous commit.

### Motivation and context

Closes #340 

https://340-spacing-tokens--mineral-ui.netlify.com/

### How to test

2. Check all of the library components for proper appearance, comparing against http://mineral-ui.com
3. Check the site components for proper appearance, comparing against http://mineral-ui.com
4. Confirm that the snapshots indicate no style changes, only theme changes

### Types of changes

Removes some theme variables.

### Checklist

Space


src/utils/styleVariables.js
```diff
@@ -58,13 +65,25 @@ export default {
...
-  spacing_double: pxToEm(16),
-  spacing_triple: pxToEm(24),
-  spacing_quad: pxToEm(32),
+  space_inline_xxs: spacing_quarter,
+  space_inline_xs: spacing_half,
+  space_inline_sm: spacing_single,
+  space_inline_md: spacing_double,
+  space_inline_lg: spacing_triple,
+  space_inline_xl: spacing_quad,
+  space_inline_xxl: pxToEm(64),
```
coldpour: why not variable-ize this one too?

kylegach: I suppose we could, but it'd be something like `spacing_octo`?

src/utils/styleVariables.js
```diff
@@ -58,13 +65,25 @@ export default {
   size_large: pxToEm(40),
   size_jumbo: pxToEm(52),
 
-  spacing_quarter: pxToEm(2),
-  spacing_half: pxToEm(4),
-  spacing_single: pxToEm(8),
-  spacing_oneAndAHalf: pxToEm(12),
-  spacing_double: pxToEm(16),
-  spacing_triple: pxToEm(24),
-  spacing_quad: pxToEm(32),
```
coldpour: how might we deprecate these vars, so that they don't disappear all at once?

src/website/app/Footer.js
```diff
@@ -26,10 +26,13 @@ const Root = createStyledComponent('div', ({ theme }) => ({
   color: theme.color_caption,
   fontSize: theme.fontSize_mouse,
   marginTop: getNormalizedValue(
-    `${parseFloat(theme.spacing_quad) * 2}em`,
+    `${parseFloat(theme.space_stack_sm) * 8}em`,
+    theme.fontSize_mouse
```
coldpour: I'm having a hard time reasoning about the changes to the inputs of getNormalizedValue. I get why we're using space_stack, but why are we using getNormalizedValue() instead of pxToEms()?

Also why are we multiplying by 8 instead of 2?

Finally, getNormalizedValue is going to parseFloat also, so we probably don't need to tack `em` on the end of line 29 or 33.

kylegach: > why are we using getNormalizedValue() instead of pxToEms()?

Because the former outputs an `em` value to be relative to the applied fontSize (`theme.fontSize_mouse`), while the latter does so relative to the base fontSize.

> Also why are we multiplying by 8 instead of 2?

Generally, if I needed an arbitrary multiplier like that, I refactored so that we're doing so on the "base unit" (8px), which is `theme.space_xx_sm`.

> Finally, getNormalizedValue is going to parseFloat also, so we probably don't need to tack em on the end of line 29 or 33.

You're right. It still expects a string, though, so I went ahead and added it for clarity.

src/utils/styleVariables.js
coldpour: spacing8x?

src/utils/styleVariables.js
kylegach: [I tried!](https://github.com/mineral-ui/mineral-ui/commit/d6a4b93dca52917c5d7004bbed2391f1b3fb0ca2#diff-7124768e5148154ecec2ba1c2f92e9ff)

Turns out, deprecating an object property is a lot more difficult than deprecating a function. I got _almost_ everything working [on this branch](https://github.com/mineral-ui/mineral-ui/blob/deprecate-theme-variables/src/utils/styleVariables.js):

- Mapping from, e.g. `spacing_single` ➡️  `space_inline_sm`, `space_inset_sm`, or `space_stack_sm`
- Deprecation message printed correctly to console as a warning
- Each deprecation message only printed once per page load

However, it would print even if none of the deprecated properties were referenced in a component. This is because I couldn't find a way to differentiate between accessing the entire object and accessing a specific property. 😕

src/utils/styleVariables.js
kylegach: Good call. I went with:

```
const spacing_quarter = pxToEm(2);
const spacing_half = pxToEm(4);
const spacing_1x = pxToEm(8);
const spacing_2x = pxToEm(16);
const spacing_3x = pxToEm(24);
const spacing_4x = pxToEm(32);
const spacing_8x = pxToEm(64);
```

src/utils/styleVariables.js
brentertz: I think that this type of change is better documented in the changelog without trying to over-engineer it.

src/website/app/Footer.js
brentertz: Agreed that `pxToEms` and `getNormalizedValue` are not very intuitive, nor is the multiplication by seemingly arbitrary values spread around everywhere. 

Perhaps the helper function names could be improved to make them more intuitive - maybe in another story?

Also, is the multiplication just to match the designs and because we don't have a spacing variable for that value?

Issue #402, opened 2017-10-09
[Dropdown] a11y announcements


brentertz:
As a mineral-ui consumer, I want the Dropdown component to have better a11y announcement support, so that I can more efficiently understand and navigate the menu using AT.

**AC:**

* 

**Other Considerations:**

* From a11y Slack:
    * Something to keep in mind for react apps is that you may want to use shouldComponentUpdate to prevent the live region from being re-rendered if the text hasn’t changed. Otherwise screen readers will keep reading it repeatedly for every render cycle

brentertz:
Recent updates to Dropdown have vastly improved a11y and keyboard navigation.  The necessity of this story needs validated ideally by someone that typically uses a screen-reader, preferably something in addition to VoiceOver.

PR #403, opened 2017-10-09
[Dropdown][Popover] followup

### LINKED ISSUE #366: [Dropdown] Followup tasks
As a Mineral UI developer, I want to complete Dropdown component related tasks that were not accomplished during the initial development

* Dropdown matches latest spec
    * Notably, top/bottom space in the menu should be removed
* A11y
    * arrow up/down does not work in voiceover mode when using keyboard to open button (when trigger is not focused)
        * Possible solution:
            * If focused trigger onOpen, aria nav would probably work
            * Probably bad for when component is controlled, but might be okay
* restoreFocus paradigm needs investigation
    * Possible solution: skip restoreFocus if click outside component
    * To illustrate, put a Popover and a Dropdown on the same page, open one, then click the trigger of the other. Notice that the one you first opened steals focus from the newly-opened one.
    * Also affects Popover

* Cross-browser issues
    * Safari/Firefox does not focus when click trigger with mouse, thus arrowUp/Down does not work
        * Possible solution: 
            * focus trigger on open
            * might also fix similar a11y issue
        * This could also be indicative that the user is using a mouse and thus it doesn't need fixed
    * Edge (maybe IE)
        * Menu open via up/down arrow does not function

* Should disabled Dropdown/Popover automatically pass disabled prop to trigger?

* Tests

brentertz:
### Description

Dropdown and Popover updates

Overview:
* Improve a11y and keyboard nav
* Simplify focus management
* Fix Dropdown isControlled pattern
* Adjust Dropdown styles per spec
* Add composeEventHandlers util
* Update examples
* Update specs

Details:
* Deletes no longer needed focus examples
* Deletes AlwaysOpenZZZ components as hiding the implementation is likely to cause more confusion than just showing all the props.  It is down to a single prop now anyway.
* Add aria-describedby to Popover/Dropdown-Trigger, referencing content element - to improve a11y announcement of content.
* PopoverTrigger now clones child and sets props on it, improving a11y trigger announcements
    * This will also now automatically disable the trigger when the Popover/Dropdown is disabled as it now passes the disabled prop to the cloned child.
* Dropdown was not fully respecting isControlled pattern
* Removes autoFocus and restoreFocus props
    * Trigger is now focused when opened and when closed
        * Call to onOpen prop follows the call to focusTrigger so user can override if desired
        * Improves arrow key navigation across browsers and voiceover/AT.
* Remove excess vertical padding from within DropdownContent

### Motivation and context

closes #366 

### How to test

https://366-dropdown-followup--mineral-ui.netlify.com/

1. View Dropdown and Popover examples
2. No more weird focus and jumping around the page when setting/focus
3. Arrow Up/Down keys can be used to open and navigate the Dropdown menu across different browsers and in voiceover
4. There is no extra padding at the top/bottom of the Dropdown menu content
5. A11y announcing is improved, though it will be further improved in #402
6. User can supply onClick/onKeyDown to Popover/Dropdown-Trigger, which will not get clobbered by Popover/Dropdown onClick, but rather they get composed

### Types of changes

Remove `autoFocus` and `restoreFocus` props from Dropdown and Popover.  Remove `DropdownContent_paddingVertical` theme variable.

### Checklist

* [ ] Rendering performance (initial load time & 60fps) and [perceived performance](http://blog.teamtreehouse.com/perceived-performance) considered


brentertz:
@coldpour - Thanks for reviewing. I added a `user-select: none` style on the DropdownContent to make the text un-selectable.

PR #412, opened 2017-10-13
fix(utils): Remove immutability-helper from createThemedComponent

brentertz:
### Description

Get rid of `immutability-helper`

### Motivation and context

Was causing an error in sample consumer app.

I like the idea of avoiding mutation, but this is the only place where the helper was being used and it feels like overkill to add a dependency just for that.  Alternatively, we could have moved it to runtime dependencies.

We'll need to publish a new release following merge...

screen shot 2017-10-13 at 12 49 25 pm

https://fix-dependency--mineral-ui.netlify.com/

### How to test

1. Try to import `createThemedComponent` in a consumer app and see if you get the error.

### Types of changes

### Checklist

* [x] Automated tests written and passing - **[n/a]**

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,6 +31,7 @@
     "no-console": ["error", {
       "allow": ["warn", "error"]
     }],
+    "no-unused-vars": ["error", {"vars": "all", "varsIgnorePattern": "[Ii]gnore"}],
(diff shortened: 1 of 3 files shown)


brentertz:
Probably not, but I'd rather get a fix published and have that talk later.

PR #414, opened 2017-10-16
[Popover][Dropdown] Portals

### LINKED ISSUE #395: [Popover] break out of scroll container
As a mineral-ui consumer, I want Dropdowns, Popovers, etc to be able to extend beyond their bounding containers even when `overflow:hidden` type styles are applied, so that I am not constrained in my designs.

AC: 

* `usePortal` prop to render popover content to body

Other considerations:

* Needed to break out of scroll container, e.g. CAAC nav menu
* keyboard nav
* a11y
* works with react 15 and 16

brentertz:
### Description

* Add `usePortal` prop to `Popover` and `Dropdown` - allows rendering of content to body instead of as sibling to trigger
* Extract reusable `Portal` component

### Motivation and context

Using a portal will allow `Popovers`, `Dropdowns`, `Tooltips`, etc to visually "break out" of bounding containers that have `overflow: hidden` styles applied.  It is needed particularly in the case where the component lives inside of a scrollable container.  

closes #395

https://395-popover-portal--mineral-ui.netlify.com/

_Unfortunately, the Netlify deploy preview appears to be caching the old version of Popper, so it doesn't look like it is working and there does not appear to be a way to clear the cache manually._

### How to test

1. Visit demo - `Popover` and `Dropdown` components
2. Note the behavior of the `Portal` examples

### Types of changes

### Checklist


brentertz:
Curious as to the team's thoughts on where `Portal` and `EventListener` components live.  Should they be siblings to the other Mineral components and simply not exposed publicly via the index?  This is how things are are currently.  Or should they perhaps live elsewhere - e.g. `/src/utils`, `/src/internal`, `/src/private`, etc.

src/Popover/Popover.js
```diff
@@ -199,21 +190,28 @@ export default class Popover extends Component<Props, State> {
       <Root {...rootProps}>
         <PopoverTrigger {...popoverTriggerProps} />
         {isOpen && popoverContent}
+        {isOpen && (
+          <EventListener
```
kylegach: 😍

kylegach:
I think keeping them directly in `/src` is fine for now. If we end up with a bunch more of these internal-only components, we can consider a better location later.

brentertz:
This PR has been stripped down to only include the Portal related updates.  Other updates were extracted and merged separately via https://github.com/mineral-ui/mineral-ui/pull/420

PR #416, opened 2017-10-16
[Website] Component status page design tweaks

riley:
### Description
- add hierarchy to pages array (and thus the navigation)
- cleanup design on Component Status page
- move some keys around to fix another markdown error

### Motivation and context
The Component Status page isn't really a Guideline page, and making the Components header a link isn't super discoverable. This PR adds a couple top-level nav headers to organize things a bit more.

Also some design tweaks so the table (with its icon-only column) doesn't look insane on larger monitors. Deprecate `margin-top` in favor of better css.

https://status-page-tweaks--mineral-ui.netlify.com/

### How to test

2. Navigate to `/component-status`
3. Check everything out. The table should feel a bit tighter.

### Types of changes
- design cleanup

### Checklist

graphic design

### DIFF
--- a/src/website/app/Router.js
+++ b/src/website/app/Router.js
@@ -22,24 +22,28 @@ import ComponentDoc from './pages/ComponentDoc';
 import ComponentDocExample from './ComponentDocExample';
 import Link from './Link';
 import LiveProvider from './LiveProvider';
-import pages from './pages';
+import sections from './pages';
 
 type Props = {
   demos: Object
 };
 
 export default function Router({ demos }: Props) {
-  const routes = [];
-
-  pages.map((page, index) => {
-    routes.push(
-      <Route
-        key={`page-${index}`}
-        path={page.path}
-        render={() => <page.component />}
-      />
-    );
-  });
+  const routes = sections
+    .map(section => section.pages)
+    // flatten array of pages arrays
+    .reduce((acc, pages) => {
+      return [...acc, ...pages];
+    }, [])
+    .map((page, index) => {
+      return (
+        <Route
+          key={`page-${index}`}
+          path={page.path}
+          render={() => <page.component />}
+        />
+      );
+    });
 
   return (
     <Switch>
--- a/src/website/app/pages/ComponentStatus/componentStatus.md
+++ b/src/website/app/pages/ComponentStatus/componentStatus.md
@@ -25,18 +25,19 @@ Statuses:
 |---|---|
 | [Button](/components/button) | [available] |
 | [Card](/components/card) | [available] |
-| DateInput | [planned] |
 | [Dropdown](/components/dropdown) ~ new | [available] |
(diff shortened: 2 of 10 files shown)


src/website/app/Router.js
```diff
@@ -22,24 +22,28 @@ import ComponentDoc from './pages/ComponentDoc';
...
-        key={`page-${index}`}
-        path={page.path}
-        render={() => <page.component />}
-      />
-    );
-  });
+  const routes = sections
+    .map(section => section.pages)
+    // flatten array of pages arrays
+    .reduce((accum, pages) => {
```
brentertz: This can be shortened.  Also, let's use `acc` for consistency.

`.reduce((acc, pages) => [...acc, ...pages], [])`

src/website/app/pages/ComponentStatus/componentStatus.md
```diff
@@ -23,20 +23,21 @@ Statuses:
 
 | Component | Status |
 |---|---|
+| NumericInput | [planned] |
```
brentertz: IMHO, available items should come before planned.  It flows better logically - **done** -> **doing** -> **will do** -> **might do**

src/website/app/pages/ComponentStatus/index.js
```diff
@@ -16,15 +16,31 @@
...
 
+const Layout = createStyledComponent(GuidelinePage, ({ theme }) => ({
+  '& table': {
+    maxWidth: '40em'
+  },
+  '& th:last-child, & td:last-child': {
+    textAlign: 'center',
+    width: '6em'
+  },
+  // this will have to be updated if there are more intro paragraphs
```
brentertz: Sentences start with capital letters, ideally even in comments.

src/website/app/pages/ComponentStatus/componentStatus.md
brentertz: Also, what is determining the order of the planned work?  For example, I would guess that we would build TextInput prior to NumericInput.

riley: That's how I had it originally, but @Mikewaite pointed out that people coming to the page probably want to see what we're working on since the "completed" components are displayed in the left nav already. Which also means that once something is "in development", it'll be at the top.

riley: I just grouped each section (planned, available, etc) alphabetically so we have some wiggle room.

brentertz: I disagree with @Mikewaite  - I think it should go in regular order, otherwise it feels like arbitrary order.

Mikewaite: That's fine. I can roll over on this. 

Done > Doing > Will Do > Might do

Issue #427, opened 2017-10-19
[Website] content updates before CA World


riley:
As a visitor to mineral-ui.com post CA world, I would like to see a polished website with most of the expected trappings.

AC:
- [x] different `<title>` element for each page [#433](https://github.com/mineral-ui/mineral-ui/pull/433)
- [x] `lang="en-US"` on the `<html>` element [#433](https://github.com/mineral-ui/mineral-ui/pull/433)
- [x] expand some acronyms e.g. RTL where needed [#436](https://github.com/mineral-ui/mineral-ui/pull/436)
- [x] better explanation of difference between `title` and `alt` attributes for Icons? [#436](https://github.com/mineral-ui/mineral-ui/pull/436)
- [x] icon category links? [#436](https://github.com/mineral-ui/mineral-ui/pull/436)
- [x] embellish copy on first Link example [#436](https://github.com/mineral-ui/mineral-ui/pull/436)
- [x] re-add example h2 above component examples [#436](https://github.com/mineral-ui/mineral-ui/pull/436)
- [x] rename all the first examples to not be "Example" [#436](https://github.com/mineral-ui/mineral-ui/pull/436)

wendyjm:
Revisit best practices and do/don't vs. always/never (if never is needed at all)

PR #433, opened 2017-10-24
[Website]: add <title> to document

riley:
connects #427 

### Description
- use React Helmet to add a unique title and canonical link to each page.
- add `lang` attribute

### Motivation and context
Since we have a single-page app, our page title never changed on navigation. This hurts SEO and accessibility. Now the title changes to give hints upon navigation.

https://document-title--mineral-ui.netlify.com/

### How to test

2. Navigate to all the pages. the format should be `title | Mineral UI`
3. `<html>` element should have a `lang` attribute set.

### Types of changes

### Checklist

open for business


src/website/app/ComponentDocExample.js
```diff
@@ -16,6 +16,8 @@
 
 /* @flow */
 import React from 'react';
+import Helmet from 'react-helmet';
+import startCase from 'lodash/startCase';
```
brentertz: I don't see `lodash.startcase` in `package.json`. Are we just able to use it because of a transitive dependency?  If so, it would probably be wise to add it explicitly.

src/website/app/ComponentDocExample.js
```diff
@@ -83,6 +89,18 @@ export default function ComponentDocExample({
 
   return (
     <Root {...rootProps}>
+      {chromeless &&
```
brentertz: Why does this care about `chromeless`?  Isn't `parent` enough? Or maybe you meant `if (!chromeless)` in order to not add titles and canonical links for `chromeless` urls?

src/website/app/Router.js
```diff
@@ -70,7 +70,11 @@ export default function Router({ demos }: Props) {
                 <IconArrowBack color="currentColor" size="small" />{' '}
                 {selectedDemo.title}
               </Link>
-              <ComponentDocExample {...selectedExample} />
+              <ComponentDocExample
+                chromeless
```
brentertz: I can't really tell what you are trying to do here with the `chromeless` prop.  This is setting `chromeless` to true for all `ComponentDocExample` child pages even when they are not `chromeless`, which I don't think is what we want.

e.g. http://0.0.0.0:8080/components/button/primary/

src/website/app/Router.js
```diff
@@ -70,7 +70,11 @@ export default function Router({ demos }: Props) {
...
+                parent={selectedDemo.title}
```
brentertz: `parent` is an overloaded term.  Can we be more explicit, e.g. `parentTitle`? 

Or even better, is it possible to calculate page title/path stuff here and pass them in rather than forcing `ComponentDocExample` to know about things?  You might need to standardize all these props to `pageTitle`, `pagePath` or something like `pageOptions: { title, path}`?  Let's discuss...

src/website/app/ComponentDocExample.js
riley: ah, i thought we had an explicit lodash dependency. updated.

src/website/app/ComponentDocExample.js
```diff
@@ -25,9 +26,14 @@ import Markdown from './Markdown';
 
 type Props = {
   backgroundColor?: string,
+  canonical?: string,
```
brentertz: No longer used...

src/website/app/Router.js
```diff
@@ -53,14 +57,24 @@ export default function Router({ demos }: Props) {
         render={route => {
           const { componentId, exampleId } = route.match.params;
           const selectedDemo = demos[componentId];
-          const selectedExample = selectedDemo.examples.filter(
+          const selectedExample = selectedDemo.examples.find(
             example => example.id === exampleId
-          )[0];
+          );
           const chromeless = route.location.search === '?chromeless';
+          const canonicalLink = `https://mineral-ui.com/components/${selectedDemo.title.toLowerCase()}/${selectedExample.id}`;
```
brentertz: You can just define `canonicalLink` and `pageTitle` in the `pageMeta` object like you do elsewhere.  No need for the extra variables if they are not used elsewhere.

```js
const pageMeta = {
  title: `${selectedDemo.title} ${startCase(selectedExample.id)} | Mineral UI`,
  canonicalLink: `https://mineral-ui.com/components/${selectedDemo.title.toLowerCase()}/${selectedExample.id}`
};
```

src/website/app/Router.js
```diff
@@ -36,11 +37,14 @@ export default function Router({ demos }: Props) {
       return [...acc, ...pages];
     }, [])
     .map((page, index) => {
+      const title = `${page.title} | Mineral UI`;
```
brentertz: You can just define `canonicalLink` and `pageTitle` in the `pageMeta` object like you do elsewhere. No need for the extra variables if they are not used elsewhere.

```js
const pageMeta = {
  title: `${page.title} | Mineral UI`,
  canonicalLink: `https://mineral-ui.com${page.path}`
};
```

src/website/app/ComponentDocExample.js
riley: whoops, missed that one.

src/website/app/Router.js
riley: no problem. the prettier formatting was hurting my eyes, so I pulled it out, but it's not a big deal.

PR #483, opened 2017-11-15
Happo demo

trotzig:
I'm using this PR to test that the happo setup is in good shape. **It shouldn't be merged**.


trotzig:
Nah, this won't work. I was hoping that this PR would be based on the last commit in the main happo PR (and not include the commits). I'll have to think of a better way to test this.

PR #497, opened 2017-11-29
[Website] add import syntax to component docs

### LINKED ISSUE #448: [Website] Add correct import syntax to Card, Menu docs
We just had a customer try to import Card components like this

`import { Card, CardTitle, CustomContent } from 'mineral-ui/Card';`

I couldn't find anywhere in the docs where we explicitly call out that `Card` is the default export.

----------

As a user, I want the import syntax for components (especially when that component is a default vs. named export) documented, so that I know how to import the components I need.

#### AC
1. Import syntax added to component pages
1. User tested with 3-ish people (expert, novice)

#### Consider
* putting under "Usage" section may be too far down the page
* put under intro paragraphs at the top of the page
* make a few deploy previews and give users a task based test

riley:
closes #448 

### Description
This PR adds documentation for import syntax per component.

### Motivation and context
There was some confusion around how/what to import from some of our components that export multiple items. Hopefully by specifically calling out the import, we can alleviate some questions.

https://448-import-syntax--mineral-ui.netlify.com/

**Other considerations**

The `Icon` syntax is a bit different since we have lots of icons. I just used `<name>` convention for the example.

I also did `import Card, {CardTitle, CardBlock, CardImage} from 'mineral-ui/Card'` for Menu and Card since this implies that these are the _only_ exports from these components. As we keep adding sub-components, e.g. `CardFooter`, we'd have to update this line. Pages for the subcomponent only show the destructured item -> `import {CardBlock} from 'mineral-ui/Card'`.

### How to test
2. Navigate to any of the component pages
3. there should be an accurate `import` example at the head of each page

### Types of changes

- feature (website upgrades)

### Checklist

make it better

### DIFF
--- a/src/website/app/Callout.js
+++ b/src/website/app/Callout.js
@@ -87,7 +87,7 @@ export default function Callout({
 }: Props) {
   return (
     <Root className={className} variant={variant}>
-      <Title variant={variant}>{title}</Title>
+      {title !== '' && <Title variant={variant}>{title}</Title>}
       {children}
     </Root>
   );
--- a/src/website/app/ComponentDocExample.js
+++ b/src/website/app/ComponentDocExample.js
@@ -35,8 +35,8 @@ type Props = {
   description?: React$Node,
   hideSource?: boolean,
   id: string,
-  scope: Object,
-  source: string,
+  scope?: Object,
+  source?: string,
   standalone?: boolean,
   title?: React$Node
 };
@@ -131,8 +131,16 @@ export default function ComponentDocExample({
     source
   };
 
+  const liveCode =
+    scope && source ? (
+      <LiveProvider
+        {...liveProviderProps}
+        chromeless={standalone && chromeless}
+      />
+    ) : null;
+
   return standalone && chromeless ? (
-    <LiveProvider {...liveProviderProps} chromeless />
+    liveCode
   ) : (
     <Root {...rootProps}>
       {standalone && (
@@ -144,7 +152,7 @@ export default function ComponentDocExample({
         {!standalone ? <Link to={id}>{title}</Link> : title}
       </Title>
       <Description scope={{ Callout }}>{description || ''}</Description>
-      <LiveProvider {...liveProviderProps} />
+      {liveCode}
     </Root>
   );
 }
--- a/src/website/app/demos/Button/examples/importSyntax.js
+++ b/src/website/app/demos/Button/examples/importSyntax.js
@@ -15,8 +15,10 @@
  */
 
 /* @flow */
-import data from './data';
-import importSyntax from './importSyntax';
-import menu from './menu';
-
-export default [importSyntax, menu, data];
+export default {
+  id: 'import-syntax',
+  title: 'Import Syntax',
+  description: `\`\`\`
+import Button from 'mineral-ui/Button';
+\`\`\``
+};
--- a/src/website/app/demos/Button/examples/index.js
+++ b/src/website/app/demos/Button/examples/index.js
@@ -15,6 +15,7 @@
  */
 
 /* @flow */
+import importSyntax from './importSyntax';
 import regular from './regular';
 import circular from './circular';
 import icons from './icons';
@@ -28,6 +29,7 @@ import states from './states';
 import truncation from './truncation';
 
 export default [
+  importSyntax,
   primary,
   regular,
   minimal,
--- a/src/website/app/demos/Card/examples/card-block/importSyntax.js
+++ b/src/website/app/demos/Card/examples/card-block/importSyntax.js
@@ -15,8 +15,10 @@
  */
 
 /* @flow */
-import data from './data';
-import importSyntax from './importSyntax';
-import menu from './menu';
-
-export default [importSyntax, menu, data];
+export default {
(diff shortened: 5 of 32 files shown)


coldpour:
Icon also has this import syntax https://448-import-syntax--mineral-ui.netlify.com/components/icon/custom/

kylegach:
> ...perhaps we should leave the syntax off the icon page since it looks to be already documented throughout the page

Or maybe replace it with a note stating to check the examples below for the import syntax? Because that's where people would expect to find that info, based on other pages.

src/website/app/demos/Button/index.js
```diff
@@ -26,6 +26,7 @@ export default {
   doc,
   examples,
   slug: 'button',
+  syntax: `import Button from 'mineral-ui/Button'`,
```
kylegach: I think `importSyntax` would be more clear, though the property's value makes it immediately obvious, so 🤷‍♂️.

src/website/app/demos/Card/index.js
```diff
@@ -51,6 +52,7 @@ Cards represent a gateway to more detailed information in another app view.`
     doc: cardBlockDoc,
     examples: cardBlockExamples,
     slug: 'card-block',
+    syntax: `import { CardBlock } from 'mineral-ui/Card'`,
```
kylegach: I wonder if subcomponents should be `import Card, { CardBlock } from 'mineral-ui/Card'`, since they're not useful outside of the default exported component? Just a thought.

src/website/app/pages/ComponentDoc/index.js
```diff
@@ -57,18 +59,40 @@ type BestPractice = {
...
+    backgroundColor: rgba(theme.color_text_primary, 0.15),
+    borderLeft: `3px solid ${theme.color_text_primary}`,
+    color: theme.color_text_primary,
+    display: 'inline-block',
+    fontFamily: theme.fontFamily_monospace,
+    fontSize: theme.fontSize_mouse,
+    lineHeight: 1.75,
+    padding: theme.space_inset_lg,
+
+    [theme.bp_moreSpacious]: {
```
kylegach: 😍 Love the much smaller text on narrow screens, where its not nearly as useful.

src/website/app/demos/Button/index.js
brentertz: I'd suggest `importSyntax` as the key as well.

brentertz:
To me, the callouts feel like they are just floating there without any context.  I'd suggest that they need an "Import Syntax" title associated with them.

brentertz:
Is there any way to get code highlighting for these snippets?

src/website/app/demos/Button/index.js
brentertz: The snippets should probably end with a `;`.

src/website/app/demos/Icon/index.js
```diff
@@ -27,6 +27,7 @@ export default {
   doc,
   examples,
   slug: 'icon',
+  syntax: `import <IconName> from 'mineral-ui-icons/<IconName>'`,
```
brentertz: You failed to document how to import the generic Icon component.

```jsx
import Icon from `mineral-ui/Icon`;
```

src/website/app/pages/ComponentDoc/index.js
```diff
@@ -57,18 +59,40 @@ type BestPractice = {
...
-})).withProps({ level: 2 });
+    [theme.bp_moreSpacious]: {
+      fontSize: theme.SiteHeading_fontSize_2_wide,
+      paddingTop: getNormalizedValue(
+        pxToEm(80),
+        theme.SiteHeading_fontSize_2_wide
+      )
+    }
+  }),
+  importSyntax: ({ theme }) => ({
```
brentertz: You called it `importSyntax` here - let's do that everywhere....

brentertz: Actually, can/should this make use of the existing `Callout` component instead?

src/website/app/demos/Button/index.js
riley: I was thinking we shouldn't have an opinion in the Great Semicolon Debate, and I did a non-scientific sample of npm READMEs, and it seemed like a mixture. I guess our codebase has `;`, so I'll update the examples.

src/website/app/demos/Icon/index.js
riley: I was wondering about that. It seemed weird to show it in the header, and using `Icon` directly is a special case that's already called out in the Custom example on that page.

src/website/app/pages/ComponentDoc/index.js
riley: yep, @kylegach and I discussed that yesterday and I'm implementing it.

src/website/app/demos/Card/index.js
riley: 👍

src/website/app/demos/Button/index.js
riley: I had that first, but I'm a sucker for brevity. `importSyntax` is probably better in the long run.

Issue #462, opened 2017-11-09
Regression testing via happo.io?


trotzig:
Hello,

This isn't an issue, it's more of a question. 

I'm building a visual regression testing service called Happo (happo.io). I'm currently in a beta phase, but things are starting to reach a good level of maturity. I've been using mineral-ui as a testing repo for a while, and I've been able to add and adjust features to make happo.io easier to work with for projects like these. 

Would you be willing to try out happo.io on your codebase? In such case, I'll put together a PR with a happo CI integration in the coming days. Once it is in place, you'll get reports like these for every PR pushed: 

<img width="498" alt="screen shot 2017-11-09 at 14 45 06" src="https://user-images.githubusercontent.com/578277/32608433-9fe6e77c-c55c-11e7-88af-7f9f616ea966.png">

You can check it out yourselves on this URL:
https://happo.io/compare?q=dev-37878da7e47ebba80258..dev-7da1818e47b2e10685db&a=14

Happo is a paid-for service, but I plan on making it free for selected open-source projects. I wouldn't charge anything for mineral-ui, as long as you provide feedback on issues you come across. Happo.io consists of three main parts. [The client](https://github.com/enduire/happo.io) is open source, but the backend service and the workers are proprietary. 

Have a great day folks!
/@trotzig

trotzig:
Ping @kylegach, I’d love to work with you folks on this.

brentertz:
Hello Henric. Thanks for reaching out! We definitely have a need for VR testing and have been wanting to do it for a while now. We are out of the office today but will review and reach out to you next week.

trotzig:
Thanks! I pushed a commit to a local branch so that you can see what a happo.io integration would look like: 
https://github.com/trotzig/mineral-ui/commit/5b5b27e9aacb1d174e8c5d760f1e21cfcf559994

What's missing from that commit is an integration with your CI environment. I'll work on that provided you're okay getting started.

brentertz:
@trotzig This looks pretty reasonable to me, and also pretty awesome!  I'm glad that using mineral as a test repo has helped you to evolve your service as well.  I'd love to see a PR so that we can try it out locally.

kylegach:
Hi, Henric! We've been heads down the last few weeks building our site, and I was going to start figuring out how to use Happo this week. How wonderful to see that you've already done so for us! 🙂

trotzig:
@kylegach I’ve noticed — things are looking good! 

[...]

PR #504, opened 2017-12-06
Testing happo integration from a mineral-ui fork

brentertz:
Testing, please ignore...

connects #462


brentertz:
@trotzig I am trying to determine the best behavior in the case where a user has forked mineral and is submitting a PR.  Due to travis security restrictions on secure keys, making them unavailable to forks, the happo script will never pass in this scenario.  I am thinking that it would likely be better to only run the happo tests if those keys are defined.  Any thoughts in this area?

brentertz:
Closing. This PR was just for testing happo integration.

PR #509, opened 2017-12-08
[Website] enable API & Theme tab for every component

### LINKED ISSUE #506: [Website] CardImage missing API subnav

riley:
closes #506 

### Description
This PR makes it so the API & Theme subnav link is _always_ shown on the component page. Also removes some unused variables.

### Motivation and context
@coldpour noticed that _only_ the CardImage page didn't have the API & Theme link, while the seemingly similar page MenuDivider did show the link. This was because MenuDivider has some theme variables, while CardImage does not. The confusing part is because at some point we made it so the API & Theme header would always show to head up the `CardImage does not have properties of its own. Undocumented properties...` callout that is _always_ shown. Now since that section is always there and contains something, it makes sense to just show the link in the subnav.

I also removed the `hidePropDoc` since we're not using it anywhere. The theory is that we were using it for the old implementation of `ThemeProvider` or something.

https://506-missing-subnav--mineral-ui.netlify.com/

### How to test

2. Click through all the component pages.
3. Every page should have an API & Theme subnav link, and content when you click that link.

### Types of changes

### Checklist

polish

### DIFF
--- a/src/website/app/pages/ComponentDoc/index.js
+++ b/src/website/app/pages/ComponentDoc/index.js
@@ -39,7 +39,6 @@ type Props = {
   description?: React$Node,
   doc: Object,
   examples?: Array<any>,
-  hidePropDoc?: boolean,
   pageMeta: {
     title: string,
     canonicalLink: string
@@ -93,7 +92,6 @@ export default function ComponentDoc({
   bestPractices,
   doc,
   examples,
-  hidePropDoc,
   componentTheme,
   title,
   whenHowToUse,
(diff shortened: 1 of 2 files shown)


src/website/app/pages/ComponentDoc/index.js
```diff
@@ -39,7 +39,6 @@ type Props = {
   description?: React$Node,
   doc: Object,
   examples?: Array<any>,
-  hidePropDoc?: boolean,
```
coldpour: I don't remember what this was for. I think @brentertz might have put it in

brentertz: As there are no longer any references to it, the need for this prop was likely removed during all of the recent website work.

PR #551, opened 2018-01-31
Theme Serializer

brentertz:
### Description

Add `themeSerializer` to exclude verbose theme output from snapshot tests.

### Motivation and context

Our snapshots are getting out of control.  This increases their value by removing some unnecessary repetition of theme output.  We don’t need to see the same theme diff on every single component.

### How to test

* Run the test suite
* Open a snapshot.  Search for `ThemeProvider`.  

### Types of changes

Update to test setup.

### Checklist


src/themes/__tests__/ThemeProvider.spec.js
```diff
@@ -22,6 +22,13 @@ import Sample from '../../website/app/demos/ThemeProvider/components/Sample';
 import examples from '../../website/app/demos/ThemeProvider/examples';
 import testDemoExamples from '../../../utils/testDemoExamples';
 
+// Allow full theme values in snapshots for ThemeProvider
```
brentertz: It would be nice to have this logic live inside `themeSerializer`, but I didn't see a good way to do so.  I'm open to suggestions or assistance.

kylegach: Agreed, but I'm not seeing a way, either.

PR #575, opened 2018-02-14
chore(webpack): add flow to webpack

coldpour:
### Description

Add the flow plugin to webpack config so that we can get flow errors in our terminals.

https://github.com/happylynx/flow-webpack-plugin
https://www.npmjs.com/package/flow-webpack-plugin

### Motivation and context

I'm using Emacs for development which doesn't have it's own special flow server like Atom does. I think this makes our DX more inclusive.

### How to test

2. make a flow error
3. see the errors in the terminal

### Types of changes

this is a chore probably

### Checklist

* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change - [n/a]

Alt text here

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,6 +20,7 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
   .BundleAnalyzerPlugin;
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const compactLicenseHeader = require('./utils/license').compactLicenseHeader;
+const FlowWebpackPlugin = require('flow-webpack-plugin');
 const NODE_ENV = process.env.NODE_ENV || 'development';
 const ANALYZE = process.env.ANALYZE;
 const isProduction = NODE_ENV === 'production';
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -69,6 +70,9 @@ module.exports = {
   devtool: isProduction ? 'source-map' : 'cheap-module-eval-source-map',
   plugins: (() => {
     let plugins = [
+      new FlowWebpackPlugin({
+        failOnError: false
```
coldpour: This was supposed to be the default, but it wasn't behaving that way. Now it's explicit. There's another config called `failOnErrorWatch` that is also supposed to be `false` by default. I'm not sure if one is better, but this seems to be working with our `npm start`.

brentertz: If we do include this plugin, we should also change the reportingSeverity to warning.  This would allow errors showing up in the console to be more quickly disregarded during a normal dev workflow.

```js
new FlowWebpackPlugin({
  failOnError: false,
  reportingSeverity: 'warning'
}),

File: webpack.config.js [hunk 2]
```diff
@@ -69,6 +70,9 @@ module.exports = {
   devtool: isProduction ? 'source-map' : 'cheap-module-eval-source-map',
   plugins: (() => {
     let plugins = [
+      new FlowWebpackPlugin({
```
brentertz: Anecdotal testing shows that the compile time doubles when making changes and saving files when this plugin is enabled.  I'd really rather not increase reload time. Perhaps this should be behind an opt-in flag, similar to `build:analyze`.  Alternatively, consider researching if there is flow integration for your editor.

**without plugin:** 1914ms, 1862ms
**with plugin:** 4090ms, 4633ms, 4861ms

brentertz:
I'm closing this issue as Mike is looking for a more localized solution that does not affect build times.

PR #577, opened 2018-02-26
build(test): Avoid posting happo comment when no diffs

trotzig:
I've noticed that the comment posted from the happo-ci.sh script can be
noisy, especially for PRs that are pushed to/rebased many times. By
silencing out the comment so that it's only posted when there are one or
more diffs/added/deleted we can avoid noise in many cases.

### DIFF
--- a/scripts/happo-ci.sh
+++ b/scripts/happo-ci.sh
@@ -68,13 +68,13 @@ run-happo "$CURRENT_SHA"
 
 # Compare reports from the two SHAs.
 COMMIT_SUBJECT="$(git show -s --format=%s)"
-SUMMARY=$(npm run --silent happo compare "$PREVIOUS_SHA" "$CURRENT_SHA" \
-  --link "$PR_URL" --message "$COMMIT_SUBJECT" || true)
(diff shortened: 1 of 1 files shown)


brentertz:
Hello @trotzig and thanks for the PR.

The Happo comments do get a bit noisy in the scenario that you mentioned.  I still believe that creating a GitHub status check would be the ideal.  This way, the Happo status would always be in one place and always up to date.  https://developer.github.com/v3/repos/statuses/

Your suggested update seems like a reasonable stop-gap.  What is lost however is the knowledge that the Happo tests actually ran successfully.

trotzig:
> The Happo comments do get a bit noisy in the scenario that you mentioned. I still believe that creating a GitHub status check would be the ideal. This way, the Happo status would always be in one place and always up to date. https://developer.github.com/v3/repos/statuses/

Agreed. It's on my radar.

> Your suggested update seems like a reasonable stop-gap. What is lost however is the knowledge that the Happo tests actually ran successfully.

Yeah, although the whole test run would fail if the happo script failed.

brentertz:
True, I suppose the build would fail in that scenario.

PR #581, opened 2018-03-07
[Select] Add new component

### LINKED ISSUE #469: [Select] component
As a user, I would like to consume a Select component that can be used on it's own, in a form or as a part of a bigger component. 

**AC**
1. works inside form field
1. update component status
1. variants (needs design: see below)
1. placeholder (null-state, unselected, other options)
1. Controlled vs. uncontrolled
1. Keyboard nav
  - Open/up/down/select/clear
1. A11y

**MVP:**
* Placeholder
* States - hover, focus, etc
* Controlled/uncontrolled
* Uses Dropdown/Popover/Menu components
* Keyboard nav
  * open/close and navigate menu
  * Simplified typeahead
* Works in FormField
* Variants
* Disabled
* ReadOnly
* Sizes

**NOT MVP:**
* Multiple item selection
* Tagging
* Input typeahead filtering
* Native select
* Avatars
* Render props
* Custom item renderer for root item

**other considerations**
* Type-ahead in the text field (auto-complete?)
* Tags (probably it's own component first)
* Multi-select (should this be another story?)
* Simple HTML select
* arrow directions

**Design Resource:**
* basic: https://ca.invisionapp.com/d/main#/console/11070131/250301744/preview

### LINKED ISSUE #582: [website] duplication on component status
Grid is so nice, we list it twice...on the component status page.

brentertz:
### Description

This PR is primarily for an MVP Select component, which provides the basic functionality of a standard HTML select component, and more.  It does not currently support searching/filtering, or multiple selection, but these are likely to be added in the future.  We are also likely to introduce render props to allow increased rendering flexibility.

* Add new Select component
* Close Select, Dropdown, Popover on blur outside component
* Adds type-ahead support to Dropdown
* Menu and Dropdown now support both flat and hierarchical data prop
* Automatic RTL based placement adjustment for Select, Dropdown, Popover, Tooltip
* Fix issue in Firefox where Dropdown immediately reopens when select item using spacebar
* Update TextArea horizontal padding to match TextInput and Select

### Motivation and context

closes #469
closes #582

http://469-select--mineral-ui.netlify.com/

### How to test

1. Open the aforementioned preview or run it locally and review the new Select component and other features described above.

### Types of changes

### Checklist


package.json
```diff
@@ -48,7 +48,8 @@
   "dependencies": {
     "exenv": "1.2.2",
     "fontfaceobserver": "2.0.13",
-    "react-popper": "0.7.4"
+    "react-popper": "0.7.4",
+    "scroll-into-view-if-needed": "1.4.0"
```
kylegach: This updated to 1.5 already.

packages/mineral-ui-icons/src/IconArrowDropdownDown.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * Copyright 2017 CA
```
Mikewaite: 2018?

src/Dropdown/Dropdown.js
```diff
@@ -147,67 +158,204 @@ export default class Dropdown extends Component<Props, State> {
...
-    return {
-      ...props,
-      'aria-activedescendant': isOpen
-        ? this.selectedItemId || `${contentId}-menu`
-        : undefined,
-      'aria-haspopup': true,
-      contentId,
-      onKeyDown: this.onTriggerKeyDown
-    };
+  getContentId = () => {
```
kylegach: Alphabetize methods?

src/Dropdown/DropdownContent.js
```diff
@@ -116,12 +118,14 @@ export default class DropdownContent extends Component<Props> {
       ...restProps
     };
 
-    const menuProps = {
-      id: `${id}-menu`,
-      data,
-      getItemProps,
-      role: 'menu'
-    };
+    const menuProps = composePropsWithGetter(
```
kylegach: Missing helpful comments.

src/Menu/Menu.js
```diff
@@ -72,8 +66,14 @@ export default function Menu({
   );
 }
 
+function isGroupedData(data) {
```
kylegach: Should this be exported and then referenced in Dropdown & Select (others)?

src/Menu/Menu.js
```diff
@@ -72,8 +66,14 @@ export default function Menu({
...
+  return data[0].hasOwnProperty('items');
+}
+
 function renderFromData(data, getItemProps) {
```
kylegach: This could be exported too.

packages/mineral-ui-icons/src/IconArrowDropdownDown.js
brentertz: Perhaps, but not right now...

package.json
brentertz: I saw that and will update.

src/Dropdown/Dropdown.js
brentertz: Maybe, but prolly not right now, unless everyone feels very strongly about it.  The methods are currently in a logical order wherein similar/related methods are grouped together.  During development, I found it much easier to work with as it avoided lots of jumping around.

src/Menu/Menu.js
brentertz: `renderFromData` logic is not duplicated anywhere that I can see.  It is similar but different from `getItems`.  Not sure there is anything to do here at present.

src/Menu/Menu.js
brentertz: Good suggestion.  I ended up moving `getItems` into Menu and importing it for use in Select/Dropdown.  This also meant that `isGroupedData` is no longer needed outside of Menu.

src/Dropdown/Dropdown.js
```diff
@@ -17,34 +17,55 @@
...
+   */
+  data: Items | ItemGroups,
+  /** @Private Function that returns props to be applied to each item */
+  getItemProps?: (props: Object, scope?: Object) => Object,
+  /** @Private Function that returns props to be applied to the menu */
+  getMenuProps?: (props: Object) => Object,
+  /** @Private Function that returns props to be applied to the trigger */
+  getTriggerProps?: (props: Object) => Object,
+  /** Index of the highlighted item. For use with controlled components. */
+  highlightedIndex?: number,
```
VictoriaVasys: It seems inconsistent for `highlightedIndex` to use an index and `selectedItem` to use an Item.

src/Select/Select.js
```diff
@@ -0,0 +1,539 @@
...
+  /** Text displayed when there is no item selected */
+  placeholder?: string,
+  /** Placement of the Select */
+  placement?: 'bottom-end' | 'bottom-start' | 'top-end' | 'top-start',
+  /** Indicates that the user cannot modify the value of the control */
+  readOnly?: boolean,
+  /** Indicates that the user must select a value before submitting a form */
+  required?: boolean,
+  /** The selected item. For use with controlled components. */
+  selectedItem?: Item,
```
VictoriaVasys: It seems inconsistent for highlightedIndex to use an index and selectedItem to use an Item.

src/Dropdown/Dropdown.js
```diff
@@ -147,67 +159,206 @@ export default class Dropdown extends Component<Props, State> {
...
-        ? acc.concat(group.items.filter(item => !item.divider))
-        : acc;
-    }, []);
+    return getItems(this.props.data);
   };
 
-  onTriggerKeyDown = (event: SyntheticEvent<>) => {
-    if (event.key === 'ArrowDown') {
+  getItemIndex = (item: Item) => {
+    return this.getItems().indexOf(item);
```
VictoriaVasys: Let's not do anything now, but keep in mind that this may cause performance issues.

brentertz: Yes, it was already noted as a followup to consider storing items as an instance variable.

src/Select/Select.js
brentertz: Good point, need to think about that one a bit...

brentertz: It is probably okay, since these are really only control props. They just mirror the state.

brentertz: We need to read values off selected item, but not the highlighted item.

src/FauxControl/FauxControl.js
```diff
@@ -0,0 +1,430 @@
...
+    ) : null;
+    const suffix = suffixIn ? (
+      <Suffix {...prefixAndSuffixProps} key="suffix">
+        {suffixIn}
+      </Suffix>
+    ) : null;
+
+    const controlProps = {
+      controlPropsIn,
+      ...controlPropsIn,
```
VictoriaVasys: Do we need to have both of these?

src/Menu/Menu.js
```diff
@@ -72,8 +66,28 @@ export default function Menu({
...
+
+const groupifyData = (data: Items | ItemGroups) => {
+  return isGroupedData(data) ? data : [{ items: data }];
+};
+
+export const getItems = (data: Items | ItemGroups) => {
+  // $FlowFixMe https://github.com/facebook/flow/issues/5885
+  const itemGroups: ItemGroups = groupifyData(data);
+  return itemGroups.reduce((acc, group) => {
+    return group.items && group.items.length
```
VictoriaVasys: Should we add a guard for group.items.filter?

src/Popover/Popover.js
```diff
@@ -280,40 +318,38 @@ export default class Popover extends Component<Props, State> {
     }
   };
 
-  isClickOutsideComponent = (event: SyntheticEvent<>) => {
+  isEventOutsideComponent = (event: SyntheticEvent<>) => {
     /* eslint-disable react/no-find-dom-node */
     const { usePortal } = this.props;
     const node = findDOMNode(this);
     const popoverContentNode = findDOMNode(this.popoverContent);
+    /* $FlowFixMe */
```
VictoriaVasys: We don't need a FlowFixMe if we use the following code: 
```js
isEventOutsideComponent = (
    event: SyntheticEvent<> | SyntheticFocusEvent<>
  ) => {
    /* eslint-disable react/no-find-dom-node */
    const { usePortal } = this.props;
    const node = findDOMNode(this);
    const popoverContentNode = findDOMNode(this.popoverContent);
    const target =
      event.type === 'blur' && event.relatedTarget
        ? event.relatedTarget
        : event.target;
```

src/Menu/Menu.js
brentertz: Filter is a method on array.prototype so it should not need guarded. We already check for items and items.length and items is flow typed to be an array, so I think it’s pretty safe.

src/Popover/Popover.js
brentertz: Cool, will try tomorrow!

src/Select/Select.js
```diff
@@ -0,0 +1,539 @@
...
+      ...restProps
+    } = this.props;
+    const isOpen = this.getControllableValue('isOpen');
+    const selectedItem = this.getControllableValue('selectedItem');
+
+    const rootProps = {
+      id: this.id,
+      ...restProps,
+      data,
+      disabled: disabled || readOnly,
```
VictoriaVasys: Why is this correct? Should there be a readOnly prop as well?

src/Select/Select.js
```diff
@@ -0,0 +1,539 @@
...
+    const { item } = scope;
+    const selectedItem = this.getControllableValue('selectedItem');
+
+    return {
+      // Props set by caller, e.g. Dropdown
+      ...props,
+
+      // Props set by this component
+      'aria-selected': selectedItem ? selectedItem.value === item.value : false,
+      onClick: this.onSelect.bind(null, item),
```
VictoriaVasys: This clobbers any `onClick` that a user might put on their items. Is that intentional or should we compose them?

src/Select/Select.js
brentertz: There is a readOnly prop on Select but not on Dropdown.  It does not make sense for Dropdown.

src/Select/Select.js
brentertz: Should we support user defined onItem click? It is easy enough to compose. The user has this capability via onChange and onSelect handlers.

src/FauxControl/FauxControl.js
brentertz: Yes.  The first is used in the application of styles and the latter is used to actually spread the props on the control.

The styles code needs `controlPropsIn.variant`, which gets overridden by `props.variant` in the definition of `controlProps`, which means that it would otherwise be unavailable to the styles code unless passed in specifically under another name.  In this case, we are just passing the entire `controlPropsIn` object for simplicity.  

There is a need to differentiate these as we may want to apply a variant inside the control, but not to the entire thing.

screen shot 2018-03-09 at 6 40 43 am

src/Select/Select.js
brentertz: I composed the functions, though I think for the majority of cases, using `onSelect` and/or `onChange` would be the preferred approach.

PR #610, opened 2018-03-13
WIP: remove what looks like unnecessary redirecting

coldpour:
can anyone remember why we have these redirects in the router? they seem unnecessary and are confusing me.

### DIFF
--- a/src/website/app/App.js
+++ b/src/website/app/App.js
@@ -119,27 +119,13 @@ class App extends Component<Props> {
     }
   }
 
-  componentDidUpdate(prevProps) {
-    if (canUseDOM && this.props.location !== prevProps.location) {
-      global.window.scrollTo(0, 0);
-    }
-  }
-
   render() {
     const { demoRoutes } = this.props;
 
     return (
       <ThemeProvider theme={siteTheme}>
         <div>
-          <Switch>
-            <Route
-              exact
-              strict
-              path="/:url*"
-              render={props => <Redirect to={`${props.location.pathname}/`} />}
-            />
-            <Route render={() => <Router demoRoutes={demoRoutes} />} />
-          </Switch>
+          <Router demoRoutes={demoRoutes} />
           <BaselineGrid />
         </div>
       </ThemeProvider>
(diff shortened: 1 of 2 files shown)


src/website/app/App.js
```diff
@@ -119,27 +119,13 @@ class App extends Component<Props> {
     }
   }
 
-  componentDidUpdate(prevProps) {
-    if (canUseDOM && this.props.location !== prevProps.location) {
-      global.window.scrollTo(0, 0);
```
kylegach: Without this one, non-Chrome browsers remembered the scroll position across page-to-page navigation.

src/website/app/Router.js
```diff
@@ -118,10 +118,6 @@ export default function Router({ demoRoutes }: Props) {
           // $FlowFixMe
           const selectedDemo = flatDemoRoutes[componentId];
 
-          if (selectedDemo.redirect) {
-            return <Redirect to={`/components/${selectedDemo.redirect}`} />;
```
kylegach: This one allows a demo to include a `redirect` property that defines which demo to redirect to.

kylegach: At the time of this comment, the Layout route uses this functionality.

src/website/app/App.js
```diff
@@ -119,27 +119,13 @@ class App extends Component<Props> {
...
 
     return (
       <ThemeProvider theme={siteTheme}>
         <div>
-          <Switch>
-            <Route
-              exact
-              strict
-              path="/:url*"
-              render={props => <Redirect to={`${props.location.pathname}/`} />}
```
kylegach: This will add a trailing slash to all routes that don't automatically have one. But as for why... 🤷‍♂️

kylegach: After some spelunking, the trailing slash was added because linking to a specific example route from a demo page couldn't be done like, e.g. `<a href="./example">`. But adding the trailing slash allowed for, e.g. `<a href="example">`.

kylegach:
Closing in favor of #612

PR #629, opened 2018-03-16
[website] Add noscript.html & noscript redirect

### LINKED ISSUE #466: [Website]: Blank screen when viewed without JavaScript
As a user visiting https://mineral-ui.com with javascript disabled, I would like it to render something besides a blank screen, so that I understand what is going on.

#### AC

* https://mineral-ui.com renders a message about requiring Javascript when viewed with JavaScript disabled
* logo, disclaimer, and link to github

#### Other considerations

* Eventually we will statically render the website and this may not be as necessary.
* Suggest to add a `<noscript />` section to `index.html` and explain that JavaScript is required to view the website.
* Consider including a "meta-refresh" to another page if anything fancy is desired. e.g.  
    ```html
    <noscript>
      <meta http-equiv="refresh" content="0;url=noscript.html">
    </noscript>
    ```

kylegach:
### Description

_See title_

### Motivation and context

Closes #466 

https://466-noscript--mineral-ui.netlify.com/noscript.html

screen shot 2018-03-16 at 12 26 23 pm

### How to test

1. Open https://466-noscript--mineral-ui.netlify.com and confirm it loads as expected
1. Disable JS and reload that page (or any other)
1. Confirm that you're redirected to https://466-noscript--mineral-ui.netlify.com/noscript.html

### Types of changes

Just website stuff.

### Checklist

You shall not pass!

### DIFF
--- a/src/website/index.html
+++ b/src/website/index.html
@@ -34,5 +34,8 @@
   </head>
   <body>
     <div id="app"></div>
+    <noscript>
+      <meta http-equiv="refresh" content="0; url=/noscript.html">
+    </noscript>
   </body>
 </html>
--- a/src/website/public/noscript.html
+++ b/src/website/public/noscript.html
@@ -1,0 +1,44 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <title>Mineral UI</title>
+    <meta charset="utf-8">
+    <meta http-equiv="x-ua-compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width">
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,600,600i,700,700i,800,800i&subset=cyrillic,greek" rel="stylesheet">
+    <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
+    <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png">
+    <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png">
+    <link rel="manifest" href="/manifest.json">
+    <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#598296">
+    <meta name="theme-color" content="#3a5663">
+    <style>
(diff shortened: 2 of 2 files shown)


brentertz:
For future reference, do all of these miscellaneous files in public need to be there or can we move them to /images or /assets if we update the paths?  The lack of organization drives me crazy.

kylegach:
I didn't even consider HTMLWebpackPlugin. Seems like overkill to me, especially for something that will likely rarely be touched.

I'm trying to mark that diff as equal and resolve the report, but neither action seems to take hold on the happo site?

They unfortunately need to be in the site root. :/

kylegach:
@trotzig — I tried signing in to happo.io with both my personal and work Google accounts. Both display the UI for resolving a report, but marking it as resolved doesn't persist and has no effect on the GitHub status. I also tried multiple browsers.

Just letting you know...

brentertz:
I think that we should at least look briefly at the HTMLWebpackPlugin as it would potentially avoid a lot of duplication.  You might also be able to import the SVG rather than copy/paste it.

kylegach:
We looked into that plugin. It seems like what it can do (take the same output bundle and inject it into two different templates) is exactly the opposite of what we need (take different outputs and inject them into the same template). Additionally, as this is a `noscript` scenario, we need to generate the full HTML content, not just a script reference to a bundle. Neither seemed trivial to figure out, so we're shipping as-is, with the possibility of optimizing later.

trotzig:
> @trotzig — I tried signing in to happo.io with both my personal and work Google accounts. Both display the UI for resolving a report, but marking it as resolved doesn't persist and has no effect on the GitHub status. I also tried multiple browsers.

Thanks for the bug report! I'm having a look at it.

trotzig:
Ah, two issues here (I think): 
- Users without admin/write access to the account are able to click the buttons (since you have a public account)
- The Accept/Reject POST 401's, but since I do an optimistic update the UI makes it look like the click persisted.

I'll work on fixing both of these issues right away. 

For reference, here are the user-access settings on your account: 
screen shot 2018-03-19 at 09 04 53

I'm guessing the "Anyone signed in with an email address on the rallydev.com domain" is problematic. Perhaps you signed in with a ca.com address? Maybe we should update the domain for your account? You can also add a invite a few email addresses directly for accounts that you may sign in as.

trotzig:
@kylegach okay, bugs are fixed - thanks for the help. Let me know if we need to do something to user access settings to simplify managing your account (adding more admins, etc).

kylegach:
Thanks, @trotzig. Sent you an email.

PR #648, opened 2018-03-28
[Text] New component

### LINKED ISSUE #158: [Text] initial component
As a consumer, I want a text component to apply the correct styles and semantics to my app's texts, so that the hierarchy is clear to all users.

**AC:**

* Component can display as a number of different text styles (inline-text, paragraphs, and headings)
    * Covers all paragraph styles from spec
* Component can be rendered as any html element in a curated list

**Other Considerations:**
* Complex (not just string) content (i.e. icons)
* Easy truncation
* Additional customization (i.e. color, font style, etc.)
- Separate Heading component built from Text?
- Variants?
- Maybe we could use `context` to dynamically determine the semantically-correct heading level?
    - This would require an app to use _all_ Mineral UI Headings

VictoriaVasys:
### Description
* New component Text
* Add Text demo & examples
* Update website content to use Text 
    * Update Markdown and Typography page
    * Did not update site headings
* Add spec & snapshots

### Motivation and context

Closes #158 

https://158-text--mineral-ui.netlify.com/components/text

### How to test

2. Navigate to Text > confirm that it matches the spec and confirm no spelling errors
3. Navigate to the Typography page & confirm it matches the spec

### Types of changes

### Checklist

Text gif


trotzig:
@VictoriaVasys can you try retriggering that Travis build? I’m hoping that the happo failure is temporary. 

(I’m the author of Happo)

kylegach:
@trotzig We've re-run Travis and it built successfully. Thanks for keeping an eye on things.

trotzig:
Great, thanks! I’ll keep monitoring this but for now let’s just hope it was a one-time thing. I haven’t been able to find a matching server error, which makes me think the 500 is coming from a proxy somewhere.

trotzig:
Sorry for the spam here again, but I just wanted to let you know that I’m working on a way for happo to inherit previous manual accepts. So that you don’t have to do the same thing ocer and over again for each new commit (assuming nothing changed in the happo diff).

kylegach:
Not spam at all. We appreciate how involved you are with our project. :)

brentertz:
This PR should should update the component status page.  
* Add the Text component with a "new" label
* Remove the "new" label from the Select component

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
+/* @flow */
+import React, { Component } from 'react';
+import { createStyledComponent } from '../styles';
+
+type Props = {
+  /** Available horizontal alignments */
+  align?: 'start' | 'end' | 'center' | 'justify',
+  /** Available styles */
+  appearance?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'mouse' | 'prose',
```
brentertz: Do we need a way for elements to be able to "appear" as normal paragraph text?

brentertz:
Nesting Text elements seems to decrease their font size.

screen shot 2018-03-30 at 6 38 04 am

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
...
+    align,
+    appearance: propAppearance,
+    color,
+    element,
+    fontWeight,
+    noMargins,
+    theme: baseTheme
+  }) => {
+    let theme = componentTheme(baseTheme);
+    const headingValues = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
```
brentertz: `headingValues` could be defined outside of this function so it is only created once.

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
...
+          return align;
+        }
+      })(),
+
+      '&::selection': {
+        backgroundColor: theme.Text_backgroundColor_selected
+      }
+    };
+
+    if (['code', 'kbd', 'samp'].indexOf(element) !== -1) {
```
brentertz: Does this list need to include `pre` as well?  It could maybe also be extracted outside of the function.

brentertz:
Have you given any thought to text style inheritance?  https://facebook.github.io/react-native/docs/text.html#limited-style-inheritance

src/website/app/demos/Text/examples/text.js
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+import IconCloud from 'mineral-ui-icons/IconCloud';
+import Text from '../../../../../library/Text';
+
+export default {
+  id: 'basic',
+  title: 'Basic Usage',
+  description: `Text renders content with the correct styles and semantics.
```
brentertz: This should probably say something about "phrasing content". 

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
...
+          return theme[`fontWeight_${fontWeight}`];
+        } else if (isHeadingAppearance && theme.Text_fontWeight) {
+          return theme.Text_fontWeight;
+        } else {
+          return fontWeight || theme.fontWeight_regular;
+        }
+      })(),
+      lineHeight: theme.Text_lineHeight,
+      // Not normalized because we actually want `##em` as applied value
+      marginBottom: !noMargins ? theme.Text_marginBottom : 0,
```
brentertz: The double negative makes this a little weird to read.  Can we reverse the order?

```js
marginBottom: noMargins ? 0 : theme.Text_marginBottom,
```

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
...
+  /** Rendered content */
+  children: React$Node,
+  /** Color of text */
+  color?: string,
+  /** The rendered HTML element, e.g. `'span'`, `'strong'` */
+  element?: string,
+  /** Available font weights */
+  fontWeight?: 'regular' | 'semiBold' | 'bold' | 'extraBold' | number,
+  /** Remove top & bottom margins */
+  noMargins?: boolean
```
brentertz: Should we offer a truncate/noWrap option?  It seems a common enough use case.

src/library/Text/Text.js
```diff
@@ -0,0 +1,165 @@
...
+  });
+}
+
+/**
+ * The Text component provides styles and semantic meaning for text and headings
+ * in a manner consistent with other components.
+ */
+export default class Text extends Component<Props> {
+  static defaultProps = {
+    align: 'start',
```
brentertz: I am not sure that we should have a default value for the align prop.  To me, it seems that not applying a default alignment would be better.  Blocks of text would then be more easily controlled by a parent layout component.

src/library/Text/Text.js
brentertz: BTW, I really like the `element/appearance` props API.
`<Text element="h1" appearance="h3">...` is super clear.

src/website/app/demos/Text/bestPractices.js
```diff
@@ -0,0 +1,95 @@
...
+          This is a paragraph. Lorem ipsum dolor sit amet, consectetur
+          adipiscing elit. Quisque molestie eros at nisl rhoncus, et condimentum
+          dui elementum.
+        </Text>
+      </div>
+    )
+  },
+  {
+    type: 'dont',
+    description: `Don't nest Text components inside one another. The styles will
```
brentertz: Can we not solve this problem?

trotzig:
@kylegach since I’m already using this PR as a discussion board, might as well ask for some more feedback: in your opinion, should a happo run merely adding new examples fail the build? It’s most likely not a failure, but it might be good to draw attention to.

src/library/Text/Text.js
kylegach: No, `pre` only implies that whitespace is preserved, not that it is displaying source code.

src/website/app/demos/Text/examples/text.js
kylegach: I don't think this is true, though. Text can be any of those linked elements, but it can also be a `p` (default), `div`, `blockquote`, and many other non-phrasing elements.

src/library/Text/Text.js
kylegach: Great suggestion.

src/library/Text/Text.js
kylegach: You can thank Victoria for that.

src/website/app/demos/Text/bestPractices.js
kylegach: We did! In addition to auto-inheriting styles when nested, we also change the element to `span` when nested in a `p` (unless a custom element is provided). And you can both opt _in_ to inheritance at the non-nested level as well as opt _out_ of it when nested.

\*phew\*

src/website/app/demos/Text/examples/text.js
brentertz: Fair enough, I guess that I was just trying to help users to create invalid DOM nesting.

src/library/Text/Text.js
brentertz: Thanks @VictoriaVasys 🙌

src/library/Text/Text.js
brentertz: It is usually displayed in a monospace font though.

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre
* https://github.com/necolas/normalize.css/blob/master/normalize.css#L57
* https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L240

src/website/app/demos/Text/bestPractices.js
brentertz: Awesome! I never doubted you for a second. 🏆

kylegach:
@brentertz — All comments addressed. See last commit.

kylegach:
@trotzig — For our workflow, no, a report with no diffs should not fail the check. I _think_ that would also make sense for most workflows, but hard to say for sure. That said, it's not a big deal to quickly affirm that it's only new snaps and approve the report.

trotzig:
@kylegach I'll hold off on making that change for now. I agree with you, the only worry I have is that examples added or deleted by mistake would go unnoticed, say someone making a fundamental change to the happo config or the local happo setup resulting in a large set of examples going missing. 

Unrelated to that, Happo will now reuse old manual resolutions (accept/reject) to a diff. If the same exact same diff appears again (which is often the case while working on a PR), the previous resolution will be reused.

src/website/app/demos/Text/examples/color.js
```diff
@@ -0,0 +1,30 @@
...
+  id: 'color',
+  title: 'Color',
+  description: `Text can be any color, but be sure to use a
+color/background-color combination with an
+[adequate contrast ratio](/color#guidelines-accessibility).`,
+  scope: { DemoPanel, palette, Text, theme },
+  source: `
+    <div>
+      <Text color={theme.color_text_success}>Success Text</Text>
+      <Text color={palette.magenta_60}>Magenta Text</Text>
```
brentertz: Suggest to rename `palette` to `colors` to match the folder name and perhaps make a better connection for the user.

src/website/app/demos/Text/examples/nested.js
```diff
@@ -0,0 +1,73 @@
...
+semantic elements by default. You can control this behavior with the \`inherit\`
+prop.
+
+<Callout title="Note">
+  <p key={0}>
+    For illustrative purposes, a purple background color and padding have been
+    applied to the Text components below.
+  </p>
+</Callout>`,
+  scope: { createStyledComponent, DemoLayout, palette, Text, theme },
```
brentertz: Again, suggest to rename `palette` to `colors` to match the folder name and perhaps make a better connection for the user.

src/website/app/demos/Text/examples/nested.js
```diff
@@ -0,0 +1,73 @@
...
+      const CustomDiv = createStyledComponent('div', {
+        backgroundColor: palette.sky_10,
+        color: palette.sky_80,
+        fontFamily: 'serif',
+        fontSize: '1.25rem',
+        padding: theme.space_inset_sm
+      });
+      return (
+        <DemoLayout>
+          For non-nested Text components, you can opt <em>in</em> to inheritance
```
brentertz: Since, "inheritance" is a loaded term, I'd suggest to precede it with the word, "style" to be explicit, as in "style inheritance".  (_Here and elsewhere..._)

src/website/app/demos/Text/examples/appearance.js
```diff
@@ -0,0 +1,24 @@
...
+import Text from '../../../../../library/Text';
+
+export default {
+  id: 'appearance',
+  title: 'Appearance',
+  description: 'Apply predefined styles with the `appearance` prop.',
+  scope: { Text },
+  source: `
+    <div>
+      <Text element="h1" appearance="h3">Heading 1 which looks like a Heading 3</Text>
```
brentertz: Not sure this is currently working as expected...

brentertz:
@kylegach I pushed a commit to fix the `contextTypes` issue without using a custom validator.  I also got it partially working using flow and the babel plugin, but it only generated the `contextTypes` property and I could not get it to create the `childContextTypes` property.  I think it is a flaw of the plugin.

Regardless, I think this is a workable solution.

PR #660, opened 2018-04-12
[Button]: filter the `element` prop from being passed down

### LINKED ISSUE #659: [Button] Styling a ReactRouter link as a Button should not throw warnings
### Expected behavior

* I should be able to style a ReactRouter link as a Button without console warnings

### Current behavior

* I have a button with an element ReactRouter link and get an error: 

```
index.js:2178 Warning: Invalid value for prop `element` on <a> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://fb.me/react-attribute-behavior
```

### Possible solution

* Add `'element'` to the `invalidLinkProps` like so:

```
function filterProps({ element, type }: Props) {
  // When element is a component, e.g. ReactRouterLink,
  // these are not filtered automatically by rootEl
  const invalidComponentProps = ['primary', 'text', 'variant'];
  const invalidLinkProps =
    element === 'a' && isTypeButton(type) ? ['type'] : ['element'];
  return Array.prototype.concat(invalidComponentProps, invalidLinkProps);
}
```

<details><summary>anything.js</summary><p>

```js
<Button primary element={ReactRouterLink} to={`/anything/`}>
  anything
</Button>
```
</p></details>

### Context

* Trying to add button styles to nav links

### Version

* "^0.26.0"

### Environment

* create-react-app v1.5.2

coldpour:
### Description

Remove the element prop from the props being passed to the DOM when Button will be rendered as an 'a' tag.

### Motivation and context

closes #659 

https://659-button-element-props--mineral-ui.netlify.com/

### How to test

1. `cd mineral-ui && npm start`.
2. Navigate to [Button > Elements](http://0.0.0.0:8080/components/button#primary) section.
3. Check your console that there are no errors/warnings.
4. Make sure the rest of the demos look laid out correctly.

### Types of changes

### Checklist

* [ ] Automated tests written and passing

huh


src/library/Button/Button.js
```diff
@@ -291,7 +291,7 @@ function filterProps({ element, type }: Props) {
   // these are not filtered automatically by rootEl
   const invalidComponentProps = ['primary', 'text', 'variant'];
   const invalidLinkProps =
-    element === 'a' && isTypeButton(type) ? ['type'] : [];
+    element === 'a' && isTypeButton(type) ? ['type'] : ['element'];
```
kylegach: I noticed that the `type` attribute is still being passed to the dom, even when the ultimately rendered HTML element is an `a`, which is invalid. I wonder if there's a way to prevent this. Maybe check for `href` or `to` props? Or only apply if an `onClick` is provided?

src/website/app/demos/Button/examples/element.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
```
kylegach: This could possibly just be a unit spec that compares snapshots, rather than a demo example. Not a big deal, either way.

brentertz: Agreed.  There is no need for this file as we never need to inspect it visually.  It can all be handled via a simple snapshot test.

src/library/Button/__tests__/Button.spec.js
```diff
@@ -1,7 +1,9 @@
 /* @flow */
 import React from 'react';
+import { BrowserRouter, Link as ReactRouterLink } from 'react-router-dom';
```
brentertz: Can we simulate a 3rd party component, without actually relying on the dependency?

src/library/Button/__tests__/Button.spec.js
```diff
@@ -29,5 +44,87 @@ describe('Button', () => {
     expect(button.props().onClick).toHaveBeenCalled();
   });
 
+  describe('types', () => {
```
brentertz: Not sure that this is the best approach...  While we are probably over reliant on snapshots, this happens to be the perfect case for one.  All of this can likely be simplified considerably.

src/library/Button/__tests__/Button.spec.js
```diff
@@ -29,5 +44,87 @@ describe('Button', () => {
...
+    describe("when it's a link", () => {
+      it(`<a> type="button" should be filtered`, () => {
```
brentertz: While I know it is relatively popular in the Ruby testing world, "should" is a superfluous word and is better avoided in "it" blocks.  That is the convention of this project anyway.  Please just use direct phrasing.  e.g. "button type is filtered", "video type is not filtered", etc.

PR #668, opened 2018-04-25
merged props are correct, I think handlers are too

coldpour:
I'm pretty sure I got the prop composition right here. I'm less certain about the event handler composition, but I think it's very close if not correct.

I left a bunch of console.logs in there for you to check out.

It turns out that the upstream/downstream didn't make sense because only one
would be defined at a time. So I went for this approach instead. Select is
always a 'listbox' now.

I thought I would have to allow one of the cases to override the event handlers,
but as far as I can tell, everything works as expected with the events as they
are now. If you can construct an experiment to prove/disprove that, that would
be awesome.


brentertz:
This is essentially the first version of the fix that I did earlier today...

src/library/Popover/Popover.js
```diff
@@ -265,40 +265,71 @@ export default class Popover extends Component<Props, State> {
...
+    /* console.log(
+     *   'Popover.getTriggerProps child.props.onBlur',
+     *   child && child.props.onBlur
+     * );
+     * console.log(
+     *   'Popover.getTriggerProps child.props.onClick',
+     *   child && child.props.onClick
+     * );
+     */
+    const triggerProps = {
```
brentertz: Yep, this looks good.  I just tightened up the syntax a bit and pushed to the main branch.  https://github.com/mineral-ui/mineral-ui/blob/9b37cf6fea29d631f0917ac815f28ebf1ecdf33e/src/library/Popover/Popover.js#L263

brentertz:
Reviewed locally and manually picked updates, with alterations, into `628-render-props` branch.

PR #671, opened 2018-04-27
test(happo): Update to happo.io 1.2.0

trotzig:
### Description

This version has a few improvements compared to 1.1.0. Those that
mineral-ui will most likely benefit from are:

- Local images (e.g. url(/foo.png)) found in CSS will now be part of the
screenshot (technically, the happo client inlines them as base64).

- No need for a customizeWebpackConfig function in config. Happo now has
a babel loader as the default config.

- Better error messages. Instead of seeing an error in a call to `eval`,
you now see the component and variant where the error happened, plus a
location in a js file (this is still the tmp file which isn't ideal, but
better than before at least).

- A polyfill for requestAnimationFrame will get rid of warnings from
React during rendering.

Full list of changes is available here:
https://github.com/enduire/happo.io/releases/tag/v1.2.0

### Motivation and context
There's mutual benefit here. Mineral-ui benefits from staying up-to-date, I make sure that happo works out in the wild. And it's a pain to support old clients out there

### How to test
Happo results for this change should show no diff or one or more diffs where images are now visible (due to better detection of local images). 

IKEA

### DIFF
--- a/.happo.js
+++ b/.happo.js
@@ -1,8 +1,6 @@
 const path = require('path');
 const { RemoteBrowserTarget } = require('happo.io');
 
-const babelLoader = require.resolve('babel-loader');
-
 const { HAPPO_KEY, HAPPO_SECRET } = process.env;
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


trotzig:
@kylegach @brentertz you might want to push this to a local branch and resubmit. I'm planning to make this whole flow work for outside PRs as well, but for now the only way to make happo trigger is if you push the branch from one of the owner/org accounts.

brentertz:
Merged via https://github.com/mineral-ui/mineral-ui/pull/672 . Thanks @trotzig

PR #687, opened 2018-05-04
[Website] Add new tokens page

brentertz:
### Description

* Add new Tokens page
* Fix Choice border styles in website
* Update Theming page table to display categorized variables
* Code gen using JS objects and JSON.stringify instead of string concatenation
* Wrap color non-string literal property keys in brackets to avoid flow warnings
* Remove private file `groupedTokens`
* Generate categorized tokens files in website for theming and tokens pages
* Remove eslint exclusion from mineral-ui-tokens `src`
* Add workflow step - run “npm run format” prior to committing generated files.

### Motivation and context

Closes #665 

### Screenshots, videos, or demo, if appropriate

https://665-tokens-page--mineral-ui.netlify.com

### How to test

1. `cd mineral-ui && npm start`
2. Navigate to Tokens; verify content & that tables render properly
3. Navigate to Theming; verify tables are organized by category & verify that theme & gray ramps display correctly

### Types of changes

- Refactor mineral-ui-tokens package
- New website page
- Breaking change - Fix incorrect sass variable name format

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support)
* [x] Automated tests written and passing - **existing coverage**
* [x] [Accessibility](http://webaim.org/intro) and [inclusivity](https://24ways.org/2016/what-the-heck-is-inclusive-design/) considered
* [x] Rendering performance (initial load time & 60fps) and [perceived performance](http://blog.teamtreehouse.com/perceived-performance) considered (n/a)
* [x] Documentation created or updated 
* [x] Tested in [sandbox](https://github.com/facebookincubator/create-react-app#getting-started) if new component or breaking change (n/a)

PR #688, opened 2018-05-04
fix(pxToEm): add tests and fix behavior of pxToEm

coldpour:
BREAKING CHANGE: pxToEm used to disregard all input units and output a value in
ems. now, it only transforms input without units or with px units. bad input
throws an error.

### Motivation and context

`pxToEm` was untested and once tested, discovered to be relatively unpredictable.

### How to test

just read the test cases :)

https://pxtoem--mineral-ui.netlify.com/

### Types of changes

### Checklist

tightening the screws

### DIFF
--- a/src/library/styles/__tests__/pxToEm.spec.js
+++ b/src/library/styles/__tests__/pxToEm.spec.js
@@ -1,0 +1,49 @@
+/* @flow */
+import { pxToEm } from '../../styles';
+
+describe('pxToEm', () => {
+  describe('converts', () => {
+    it('integers', () => {
+      expect(pxToEm(2)).toEqual('0.125em');
+    });
+
+    it('floats', () => {
+      expect(pxToEm(2.5)).toEqual('0.15625em');
+    });
+
+    it('strings without units', () => {
+      expect(pxToEm('8')).toEqual('0.5em');
+    });
+
+    it('strings with px units', () => {
+      expect(pxToEm('4px')).toEqual('0.25em');
+    });
+
+    it('strings with floats', () => {
+      expect(pxToEm('2.5')).toEqual('0.15625em');
+    });
+
+    it('strings with floats and px units', () => {
+      expect(pxToEm('2.5px')).toEqual('0.15625em');
+    });
+  });
+
+  describe('errors with bad input', () => {
+    [
+      '2em',
+      '2rem',
+      '',
+      'string without numbers',
+      'px',
+      {},
+      ['array'],
+      () => {}
+    ].forEach((input) => {
+      // $FlowFixMe: We're fine with coercing the input here
+      it(`${input}`, () => {
+        // $FlowFixMe: We want the error
+        expect(() => pxToEm(input)).toThrowError();
+      });
+    });
+  });
+});
--- a/src/library/styles/pxToEm.js
+++ b/src/library/styles/pxToEm.js
@@ -4,8 +4,22 @@
  */
 import fontSize_base from '../themes/fontSizeBase';
 
+const REGEX_NUM_OR_END_IN_PX = /^[\d.]+(px)?$/;
+
+const convert = (input: number): string =>
+  `${input / parseInt(fontSize_base)}em`;
+
+const errorMsg = (actual) => {
+  const value = typeof actual === 'string' ? actual : typeof actual;
+  return `[mineral-ui/styles/pxToEm]: Expected a number or string in px units. Instead got: '${value}'.`;
+};
+
 export default function pxToEm(value: number | string) {
-  const input =
-    typeof value === 'string' ? parseInt(value.replace('px', '')) : value;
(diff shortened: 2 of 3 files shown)


src/library/styles/__tests__/pxToEm.spec.js
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    it('respects floats', () => {
+      expect(pxToEm(2.5)).toEqual('0.15625em');
+    });
+  });
+
+  describe('with other input types', () => {
+    it('errors', () => {
+      // $FlowFixMe
```
kylegach: You should probably add a brief explanation (e.g. "We want this error") as to why these are necessary.

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,26 @@
...
-  return `${input / parseInt(fontSize_base)}em`;
+  switch (typeof value) {
+    case 'number':
+      return convert(value);
+      break;
+
+    case 'string':
+      return justNumbersOrEndsWithPx(value)
+        ? convert(parseFloat(value))
+        : value;
```
kylegach: I thought this case would throw an error as well.

kylegach:
Also, relatively minor, but our `BREAKING CHANGE`s have been in the imperative tense like our commit summaries and use proper sentence structure.

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,20 @@
  */
 import fontSize_base from '../themes/fontSizeBase';
```
brentertz: Related: `fontSize_base` is just doing a replace on `px`.  Shouldn't it just use `parseFloat`?

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,20 @@
...
 
+const justNumbersOrEndsWithPx = (s: string) => s.match(/^[\d.]+(px)?$/);
```
brentertz: * Regex should be extracted to const so it does not get recreated each time this is called.
```js
const REGEX_NUM_OR_PX = /^[\d.]+(px)?$/;
```
* Can use `test` rather than `match` if you are not doing anything with the matches.
* Actually this function is really unnecessary as the `test` can simply be inlined.
```js
REGEX_NUM_OR_PX.test(value)
```

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,20 @@
...
+
+const convert = (input: number): string =>
+  `${input / parseInt(fontSize_base)}em`;
```
brentertz: This should prolly use `parseFloat` as well, no?

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,20 @@
...
+
+const coerce = (value) => (typeof value === 'string' ? value : typeof value);
```
brentertz: `coerce` could prolly be named better - `valueOrType` perhaps?  Also, it probably does not need to be a separate function if it is only called once.  Consider just creating a const inside the error function.

src/library/styles/pxToEm.js
```diff
@@ -4,8 +4,20 @@
...
+const errorMsg = (actual) =>
+  `[mineral-ui/styles/pxToEm]: Expected a number or string in px units. Instead got: '${coerce(
+    actual
+  )}'.`;
+
 export default function pxToEm(value: number | string) {
-  const input =
-    typeof value === 'string' ? parseInt(value.replace('px', '')) : value;
-  return `${input / parseInt(fontSize_base)}em`;
+  if (typeof value === 'number') return convert(value);
```
brentertz: Not sure why these are not eslint `curly` errors.  We should probably always use braces here. 
 https://eslint.org/docs/rules/curly

Suggest to add `"curly": ["error"]` to `.eslintrc` rules section.

src/library/styles/__tests__/pxToEm.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      'string without numbers',
+      'px',
+      // $FlowFixMe
+      {},
+      // $FlowFixMe
+      ['array'],
+      // $FlowFixMe
+      () => {}
+    ].forEach((c) => {
+      it(`${c}`, () => {
```
brentertz: This probably doesn't need to generate a separate `it` for each value, but rather could be a single `it` that tests multiple values.

brentertz:
The commit message "scope" should be `styles` rather than `pxToEm`.

src/library/styles/pxToEm.js
brentertz: Oh nevermind, this update is coming in #687

src/library/styles/pxToEm.js
kylegach: `fontSize_base` will always be an integer value, as the entire system is built upon it, and we want easy math.

PR #699, opened 2018-05-24
[Website] Refactor DemoLayout

VictoriaVasys:
### Description
* Add shared component `DemoLayout` and use where appropriate in site demos

### Motivation and context
* Reduces much repetition among site demos

https://shared-demo-layout--mineral-ui.netlify.com/

### How to test

1. Navigate to Avatar, Button, Checkbox, FormField, Link, Radio, TextArea & TextInput and ensure no changes

### Types of changes
- Refactor

### Checklist


brentertz:
I like where you are headed - reducing code duplication.

I think that one of the reasons why this hasn't been done yet is because doing so muddies the simplicity of the demo code examples.  How can we get closer to the first one?  I think it is important to minimize distraction here if possible.

```js
<DemoLayout></DemoLayout>
```

vs.

```js
<DemoLayout includeLastChild marginRight="1rem"></DemoLayout>
```

A couple of things to consider...

Can we use a Box component as the root of DemoLayout?  We'd then get theme correct spacing props for free.  We could set default props, so they wouldn't have to be displayed in the code example.  Can we make all of the spacing/margin consistent across all demos so that we don't typically need to provide any props to this component?  Can we normalize the last-child css so that it is consistent in all of our demos and avoid the prop altogether?

VictoriaVasys:
@brentertz Good point, it does add unnecessary noise to the site. Consulted with @kylegach on the Box approach & since these styles apply to children, it doesn't make sense to go that route. Unfortunately, we can't normalize the last-child one because some displays are side-by-side, where you must style the last item's margin-bottom, whereas some are displayed in a vertical manner, where a margin-bottom wouldn't be appropriate for the last child. 

Kyle came up with a clever way around it though, so now there is no change to how these are displayed in examples.

PR #701, opened 2018-05-24
Update Functional Components to Fix Flow-to-Prop-Type Conversion

VictoriaVasys:
### Description
* `flow-to-prop-types` babel plugin does not allow props to be destructured in function declaration (Flow errors were not rendered to the console)

### Motivation and context
* Flow errors are now properly rendered

https://functional-prop-types--mineral-ui.netlify.com/

### How to test

1. Pass a prop with an unaccepted prop type to Avatar, CardActions, CardBLock, CardStatus, CardTitleMenu, Checkbox, CheckboxGroup, Choice, ChoiceGroup, Flex, FlexItem, FormFieldset, Grid, Link, MenuGroup, Radio, RadioGroup, TextInput or ThemeProvider & you should receive a console error.
2. Check any of the above component pages to ensure that correct default props are documented

### Types of changes

### Checklist
* [x] Rendering performance (initial load time & 60fps) and [perceived performance] [n/a](http://blog.teamtreehouse.com/perceived-performance) considered

Dog on flamingo, flowing down a river


src/website/app/Nav.js
```diff
@@ -17,7 +17,7 @@ import sections from './pages';
 type Props = {
   currentDemo?: string,
   demoRoutes: Array<DemoRoute>,
-  wide?: Boolean
+  wide?: boolean
```
brentertz: Nice catch!

I am getting console errors when running this branch locally though...

----
Warning: Failed prop type: Invalid prop `wide` of type `Boolean` supplied to `Nav`, expected instance of `Boolean`.

Warning: Failed prop type: The prop `pageMeta` is marked as required in `ComponentDoc`, but its value is `undefined`.

src/library/Avatar/Avatar.js
```diff
@@ -134,10 +131,14 @@ export default function Avatar({
...
     noBackground,
-    shape,
     size,
     ...restProps
   };
 
   return <Root {...rootProps}>{text || icon || children}</Root>;
 }
+
+Avatar.defaultProps = {
```
brentertz: It is okay to define the defaultProps after the export?  It seems to be working...

src/website/app/Nav.js
brentertz: Hmmm, I think I was on an out of date version...  I updated and these errors went away.

src/library/Avatar/Avatar.js
VictoriaVasys: @kylegach & I were both wondering that. There were a few examples of `displayName` being defined after the export, so I experimented with `defaultProps` & although it seems to work, our inclination is that maybe it's because of a transpile & that may not be safe for our users? So I changed them all back to export after the assignments. 

However, in `Avatar` & `Link`, where `size` & `element` are optional props (respectively, they error when the defaults are defined outside of the declaration. We got around this by accessing the default props, e.g.:

```js
const Link = (props: Props) => {
  const Root = createStyledComponent(
    element || this.defaultProps.element,
    linkStyles,
    ...
  )
  ...
}
Link.defaultProps = {...}
export default Link;
```

This was fine when we had `export default function <Component>`, but Flow errors with the above pattern. For now, I changed those two components back, but we should discuss what the best pattern is & whether the above approach makes sense.

VictoriaVasys: Disregard that, @coldpour helped me realize that it was just a matter of `this` changing with the arrow function. Fixed it & now all exports are at the ends of files.

brentertz: Yeah, probably not a big deal either way...since it works regardless...

src/library/Icon/IconArrowDropdownDown.js
```diff
@@ -10,7 +10,7 @@ type Props = {
 };
 
 /* eslint-disable prettier/prettier */
-export default function IconArrowDropdownDown(props: Props) {
+const IconArrowDropdownDown = (props: Props) => {
```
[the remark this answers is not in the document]
VictoriaVasys: Sounds good- thanks for catching that. I didn't realize they were generated.

PR #708, opened 2018-06-06
Migrate from Glamorous to Emotion

### LINKED ISSUE #692: Emotion migration
_The Glamorous project has been deprecated and the suggested path forward is to migrate to Emotion._

https://github.com/paypal/glamorous/issues/419

### Initial investigation

Emotion "may" be able to do all of the things that Glamorous can do, but not with the same API and not out of the box.

**API Docs**
  * [Glamorous](https://glamorous.rocks/api/)
  * [Emotion](https://emotion.sh/docs)
    * [Stylis](https://stylis.js.org/) - Emotion uses this internally

**Performance**
  * Emotion is reported to be both smaller and faster than Glamorous
  * Here are some CSS-IN-JS benchmarks - https://github.com/A-gambit/CSS-IN-JS-Benchmarks/blob/master/RESULT.md

----
**AC:**

* [x] Site renders properly and without warnings
* [x] Components render properly and without warnings
* [x] Migration documentation created for consumers

**Tasks:**

* [x] Remove Glamorous dependencies
  * [x] `glamor` is needed as a dev dependency until the `simulate` helper usages have been removed
  * [x] `glamorous`, `jest-glamor-react`
  * [x] `css` prop is not compatible with `babel-plugin-transform-react-inline-elements`.  We have opted to NOT use `babel-plugin-emotion`, so this can remain intact.
* [x] Add new Emotion dependencies
  * `emotion`, `react-emotion`, `emotion-theming`, `jest-emotion`, `eslint-plugin-emotion`, `recompose`
* [x] Update `createThemedComponent`
* [x] Update `createStyledComponent`
  * [x] Css label identifiers in development only
  * [x] Verify preset filtered props
* [x] Update `ThemeProvider`
* [x] Update `withTheme`
* [x] Implement replacement for glamorous `withProps`.
* [x] Configure `jest-emotion` snapshot serializer
* [x] Update install documentation
* [x] Update examples and other documentation
* [x] Remove or reimplement glamor `simulate`.  See https://github.com/emotion-js/emotion/issues/666 This has been removed for now.
* [x] [Emotion uses WeakMap](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L166) so we would potentially need to document the need for users to polyfill this in their apps.  WeakMap is mostly supported in IE11 though and as far as I can tell by their usage, it should work.
  * Tested website in IE11 VM with polyfills disabled via `useBuiltins: false` in `babelrc.js`.  No problemos.
* [x] Drop support for `css` prop until Emotion implements it without a babel plugin.
* [x] There are a couple places in the website that use glamor `css` directly.  `import { css } from 'glamor';` These need updated. - `Loading.js` and `NotFound.js`
* [x] React 16.4 adds support for pointer events - add them to `reactProps.js`
* [x] Can remove no longer needed "inert" styles inside of several createStyledComponent calls that were used to circumvent glamorous cache issue.  These are difficult to search for since the styles used may not always be the same.  They can be removed incrementally in the future as we come across them.  Mineral developers should note that they are no longer needed.
* [x] Update to new version of Emotion containing our `shouldForwardProp` composition fix, once released.

brentertz:
### Description

Migrate Mineral UI's underlying CSS-in-JS framework from Glamorous (deprecated) to Emotion

* Updates mineral dependencies, API, components, website, and docs to use Emotion instead of Glamorous
* Adds [Emotion migration guide](https://github.com/mineral-ui/mineral-ui/blob/692-emotion/docs/emotion-migration.md) - Please read it to become familiar with the changes.
* Removes states examples as there is no way to simulate CSS pseudo classes 
* Updates flow-typed

### Motivation and context

closes #692 

https://692-emotion--mineral-ui.netlify.com/

### How to test

1. Review website and validate that things look the same
1. Review website content, particularly the styling page - regarding `createStyledComponent` and using `className`
1. Review install docs
1. Validate that there are no visual regressions
1. Review snapshot diffs
1. I have published a prerelease alpha version which you can use to try out in your own app if you like.  Install via `npm i mineral-ui@alpha` 

### Types of changes

### Checklist

emotional


src/library/Select/SelectTrigger.js
```diff
@@ -89,13 +90,15 @@ const styles = {
...
+      // the arrow icon
       '& :not([role="img"]) ~ [role="img"]': {
         color:
           disabled || readOnly
             ? theme.color_disabled
             : variant ? theme[`icon_color_${variant}`] : theme.SelectIcon_color
       },
 
+      // the variant icon
       '& :not([role="img"]) + [role="img"]:not(:last-of-type)': {
```
kylegach: What... you can't parse that selector quickly enough? 😝

src/library/styles/__tests__/createStyledComponent.spec.js
```diff
@@ -0,0 +1,121 @@
...
+      { filterMe: true },
+      {},
+      {
+        filterProps: ['filterMe']
+      }
+    );
+    const button = wrapper.childAt(0);
+
+    it('filters filterProps', () => {
+      expect(button.props().filterMe).toBeUndefined();
```
kylegach: Is this test & expectation testing something different than the one on line 16?

Or, put another way, is testing based on `.props()` sufficient to ensure something is actually in the DOM?

kylegach: I may be misunderstanding, but I don't understand how this assertion is meaningfully different from the one on line 20. Wouldn't you need to filter a valid DOM attribute for this to be useful? Or, put another way, is asserting on `.props()` an accurate representation of what's actually in the DOM?

src/library/styles/__tests__/createStyledComponent.spec.js
```diff
@@ -0,0 +1,121 @@
...
+      { forwardMe: true },
+      {},
+      {
+        forwardProps: ['forwardMe']
+      }
+    );
+    const button = wrapper.childAt(0);
+
+    it('forwards forwardMe', () => {
+      expect(button.props().forwardMe).toBeDefined();
```
kylegach: I may be misunderstanding, but I don't understand how this assertion is meaningfully different from the one on line 16. Wouldn't you need to forward a non-valid DOM attribute for this to be useful? Or, put another way, is asserting on `.props()` an accurate representation of what's actually in the DOM?

src/library/styles/createStyledComponent.js
```diff
@@ -1,31 +1,76 @@
 /* @flow */
-import glamorous from 'glamorous';
+
+import styled from 'react-emotion';
+import withPropsFn from 'recompose/withProps';
 import componentStyleReset from './componentStyleReset';
+import isValidProp from '../utils/isValidProp';
+import { isProduction } from '../utils/nodeEnv';
 
 export default function createStyledComponent(
```
kylegach: This is just preference, but I would find this easier to read:

```js
type Element =
  | React$StatelessFunctionalComponent<*>
  | React$ComponentType<*>
  | string;
type Styles =
  | Object
  | Array<Object>
  | ((props: Object, context?: Object) => Object | Array<Object>);
type Options = {
  displayName?: string,
  filterProps?: Array<string>,
  forwardProps?: Array<string>,
  includeStyleReset?: boolean,
  rootEl?: string,
  withProps?: Object
}

export default function createStyledComponent(
  element: Element,
  styles: Styles
  options?: Options  = {}
) {
```

src/library/styles/createStyledComponent.js
```diff
@@ -1,31 +1,76 @@
...
-  const { includeStyleReset, ...restOptions } = options;
-  let outStyles;
+  const {
+    displayName,
+    filterProps = [],
+    forwardProps = [],
+    rootEl,
+    withProps
+  } = options;
+  const includeStyleReset = options.includeStyleReset || false;
```
kylegach: I assume flow required this? Curious as to why...

src/library/styles/createStyledComponent.js
```diff
@@ -1,31 +1,76 @@
...
+    props: Object,
+    context?: Object
+  ): Object | Array<Object> => {
+    let componentStyles =
+      typeof styles === 'function' ? styles(props, context) : styles;
 
-      return {
-        ...componentStyleReset(props),
+    if (Array.isArray(componentStyles)) {
+      componentStyles.unshift(componentStyleReset(props));
```
kylegach: If `styles` is an array, `componentStyleReset` is always included?

src/library/styles/createStyledComponent.js
```diff
@@ -1,31 +1,76 @@
...
+  };
+
+  if (displayName && typeof element !== 'string') {
+    element.displayName = displayName;
   }
 
-  return glamorous(element, restOptions)(outStyles);
+  const styledComponent = styled(element, {
+    ...(!isProduction && displayName ? { label: displayName } : undefined),
+    shouldForwardProp: (prop) => {
```
kylegach: I know there's a lot of learning embedded in this property, but I found it super-clear to read and understand. 👏

src/library/themes/ThemeProvider.js
```diff
@@ -1,22 +1,28 @@
...
- * ThemeProvider provides a theme to the tree of components contained within.  See the [theming page](/theming) for more information.
+ * ThemeProvider provides a theme to the tree of components contained within.
+ * See the [theming page](/theming) for more information.
  */
 const ThemeProvider = (props: Props) => {
   const { children, theme } = props;
   return (
-    <GlamorousThemeProvider theme={theme}>{children}</GlamorousThemeProvider>
+    <EmotionThemeProvider theme={theme}>
+      {Children.only(children)}
```
kylegach: Nice. This should produce more helpful errors.

src/library/themes/ThemeProvider.js
```diff
@@ -1,22 +1,28 @@
...
 type Props = {
   /** Components to which the theme will be applied */
   children?: React$Node,
-  /** A shallow object of [theme variables](/theming#common-scenarios-theme-structure) and their values */
-  theme?: Object
+  /**
+   * A shallow object of [theme variables](/theming#common-scenarios-theme-structure)
+   * and their values or a function that provides such an object.
+   */
+  theme?: Object | (() => Object)
```
kylegach: Was this always the case and we typed it incorrectly? Or is this new functionality? Either way, 👍.

src/library/utils/isValidProp.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
+
+import _htmlAttributes from 'react-html-attributes';
+import memoize from 'fast-memoize';
+import reactProps from './reactProps';
+import { settify } from './collections';
+
+type HtmlAttributes = {
+  '*'?: Set<string>,
```
kylegach: I didn't know you could do that! That would've come in handy in the `createTheme` work a few PRs back...

kylegach: Or maybe I'm misunderstanding. Is that literally an object property with a key of `'*'`?

docs/emotion-migration.md
```diff
@@ -0,0 +1,51 @@
+# Emotion Migration Guide #
```
kylegach: This is really great!

src/website/app/Page.js
```diff
@@ -244,13 +244,15 @@ const styles = {
         paddingTop: pxToEm(53), // roughly theme.baseline_5
 
         '&::before': {
+          content: '""',
```
kylegach: Header in this file is a styled Section, which already has this property for these selectors: https://github.com/mineral-ui/mineral-ui/blob/master/src/website/app/Section.js#L59

src/website/app/demos/Flex/bestPractices.js
```diff
@@ -72,17 +73,17 @@ align components to a columnar layout. Use [Grid](/components/grid) and
       description: `Don't use Flex if the content being laid out should _not_
 [respond to language direction](#rtl).`,
       example: (
-        <Flex css={{ height: 200 }}>
+        <Flex className={css({ height: 200 })}>
```
kylegach: Flex (which I wrote) has a `height` prop, so I have no idea why this (which I wrote) was using `css` here. 🤦‍♂️

src/website/app/demos/Flex/examples/Flex/alignItems.js
```diff
@@ -1,10 +1,11 @@
 /* @flow */
+import withProps from 'recompose/withProps';
 import { createStyledComponent } from '../../../../../../library/styles';
 import FlexItem from '../../components/FlexItem';
 import _DemoLayout from '../../components/DemoLayout';
 import _Flex from '../../components/Flex';
 
-const DemoLayout = _DemoLayout.withProps({ lastRowStartsAt: 5 });
```
kylegach: Ahh, this is probably why glamorous made it a method instead of an option. Good to know.

src/website/app/demos/Grid/bestPractices.js
```diff
@@ -36,16 +37,16 @@ components to a columnar layout.`,
       description: `Don't use Grid if the content being laid out should _not_
 [respond to language direction](#rtl).`,
       example: (
-        <Grid css={{ height: 200 }}>
+        <Grid className={css({ height: 200 })}>
```
kylegach: Same as Flex's best practices; please change to use `height` prop (sorry).

src/website/app/demos/Link/examples/variants.js
```diff
@@ -1,7 +1,14 @@
 /* @flow */
 import React from 'react';
+import { createStyledComponent } from '../../../../../library/styles';
 import _Link from '../../../../../library/Link';
-import { DemoLayout } from './states';
+import _DemoLayout from '../components/DemoLayout';
+
+const DemoLayout = createStyledComponent(_DemoLayout, {
```
kylegach: I think you could use the [shared DemoLayout](https://github.com/mineral-ui/mineral-ui/blob/master/src/website/app/demos/shared/DemoLayout.js) Victoria made instead.

src/library/Avatar/__tests__/__snapshots__/Avatar.spec.js.snap
```diff
@@ -1,192 +1,171 @@
...
+  font-family: "Open Sans",-apple-system,system-ui,BlinkMacSystemFont,"Segoe UI",Roboto,Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol";
   font-size: 16px;
   line-height: 2.5em;
   outline: 0;
   -moz-osx-font-smoothing: auto;
   -webkit-font-smoothing: antialiased;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
+  -ms-flex-align: center;
   align-items: center;
```
kylegach: It makes me inordinately happy that vendor prefixed properties are now grouped with the non-vendor-prefixed one and, more importantly, are printed first so that the non-vendor-prefixed one wins if supported. Thanks, emotion! 👩‍🎤

src/library/utils/isValidProp.js
coldpour: yes literally a `'*'` key in the object. it's where all the attributes that apply to every DOM element live in that import

src/library/themes/ThemeProvider.js
brentertz: This is new with emotion.

src/website/app/Page.js
brentertz: Removed... Must've added when trying to get site to render initially.

src/website/app/demos/Flex/bestPractices.js
brentertz: Fixed.

src/website/app/demos/Flex/examples/Flex/alignItems.js
brentertz: Well, one could also just write the following...

```js
const DemoLayout = createStyledComponent(_DemoLayout, {}, { lastRowStartsAt: 5 });
```

src/website/app/demos/Grid/bestPractices.js
brentertz: Fixed, no problem...

src/website/app/demos/Link/examples/variants.js
brentertz: Yep.

src/library/styles/createStyledComponent.js
brentertz: Wops! Fixing...

src/library/styles/createStyledComponent.js
brentertz: Nope...left from initial debuggerating... Fixed, thanks!

src/library/styles/__tests__/createStyledComponent.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    it('filters invalid DOM attributes', () => {
+      // src is invalid here because rootEl is div
+      expect(image.props().src).toBeUndefined();
+    });
+  });
+
+  describe('styles', () => {
+    const styleObject = { color: 'red' };
+    const styleArray = [{ color: 'red' }, { backgroundColor: 'blue' }];
+    const styleObjectFunction = () => styleObject;
```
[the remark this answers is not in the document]
brentertz: Perhaps, but I think that we would know pretty quickly if we broke that aspect.  e.g. Wouldn't just about everything break in appearance?  This group of tests were added while diagnosing that we didn't support style arrays.

src/library/styles/__tests__/createStyledComponent.spec.js
brentertz: `forwardMe` is an invalid DOM attribute.  Line 16 asserts on a valid DOM attribute.

Regarding it being valid to test `.props()`, that is a good question.  I think so since, commenting out the forwardProps option causes the test to fail.  FWIW, I lowercased the forwardme prop and set the value to a string, which gets rid of the console warning and also causes the attribute to show when logging button.html().  I could also assert on the button.html() string, but not sure if it is worth it as this seems to already be testing the mineral part of forwardProps functionality.  I think a snapshot would essentially be the same as it is currently, just less explicit.  Open to other suggestions…

src/library/styles/__tests__/createStyledComponent.spec.js
brentertz: Agreed, it is unclear if the prop is being filtered via the `filterProps` setting or via the DOM tag.  It would work if it were styling a Button component and not a button tag.  Regardless, I updated this to filter a valid DOM attribute.

src/website/app/demos/Flex/examples/Flex/alignItems.js
kylegach: For internal or site-only work, either is fine by me. For public-facing code, we should only use `createStyledComponent`.

brentertz: For consistency, I just updated them all to use the `createStyledComponent` instead of `withProps`.

src/website/app/pages/Styling/styling.md
```diff
@@ -57,14 +66,18 @@ This function is used to create a new styled component based on another componen
...
+
+| Option              | Type          | Description                                                                                        |
+|---------------------|---------------|----------------------------------------------------------------------------------------------------|
+| `displayName`       | String        | Sets component displayName and CSS className label during development                              |
+| `filterProps`       | Array<String> | Props that should not be passed to child components                                                |
+| `forwardProps`      | Array<String> | Props that should be passed to child components. Takes precedence over `filterProps` and `rootEl`. |
+| `includeStyleReset` | boolean       | Includes a minimal CSS style reset                                                                 |
+| `rootEl`            | String        | HTML tag name of the rendered DOM node. Helps ensure that only valid attributes reach the DOM.     |
+| `withProps`         | Object        | <key, value> pairs of props to apply to a component                                                |
```
VictoriaVasys: This table makes it much easier to identify/find what's possible in the options object. Nicely done!

PR #719, opened 2018-06-18
test(happo): Run happo on PRs from forks

trotzig:
### Description

Since happo.io@1.5.0 (which is the version used on mineral-ui master),
`apiKey` and `apiSecret` are no longer needed for builds triggered by a
PR. Instead, a separate auth mechanism kicks in to allow the happo run
to succeed. The authentication used here is limited in scope (only
allowed on specific commit shas) and in time (has a short ttl).

In most scenarios, the limited authentication is enough. There are
edge-cases however, where the happo run might fail unexpectedly. Some of
the cases where this can happen:
- Re-triggering a happo build when no code change has been made to the
PR
- Long wait times from CI causing tokens to time out.

A pre-requisite for token-less authentication to work is that the Happo
Github App is installed and properly configured on
https://happo.io/github-integration (I've already done this, so no need
for action here).

### Motivation and context

Developers who are unfamiliar with the codebase are usually the ones needing screenshot testing the most. It's easy to mess up styling when you don't have enough context. Running Happo on outside contributors will make it easier to contribute. 

### How to test

If my setup works, you should see Happo post a result (no diffs) on this PR. If things aren't working, we'll likely see a failure in the Travis run. It's also possible, but unlikely, that things silently fail. We'll have to make sure to check the travis logs.


trotzig:
That was the result (no diffs) I was hoping for 😊

brentertz:
Thanks @trotzig

PR #750, opened 2018-07-29
[Flex,FlexItem] Flex gutter refactor & improvements; FlexItem can be a Flex

### LINKED ISSUE #706: [Flex] FlexItem children don't properly support RTL
### Expected behavior

What doesn't work below should work.

### Current behavior

Passing `marginStart` or `marginEnd` on a FlexItem child should override the gutter gap; it does not. Also, Flex is applying `marginLeft` or `marginRight` with no respect to RTL (which I _think_ matters?)

### Possible solution

Not sure yet, but it'll involve these lines: https://github.com/mineral-ui/mineral-ui/blob/master/src/library/Flex/Flex.js#L187-L188

1. Render the code below somewhere
2. In an LTR environment, A should be on the left side of the container and B on the right
3. Instead, B currently sits immediately to the right of A, a gutter gap apart
4. In a RTL environment, A should be on the right side and B on the left
5. Instead, B currently sits immediately to the left of A, a gutter gap apart

<details><summary>RenderThisSomewhere.js</summary><p>

```jsx
<Flex>
  <FlexItem>A</FlexItem>
  <FlexItem marginStart="auto">B</FlexItem>
</Flex>
```
</p></details>

### Context

Prevents overriding Flex's gutter gap with per-FlexItem margins.

### Version

v0.33.0

### LINKED ISSUE #710: [StartEnd] doesn't go to start in IE11
### Expected behavior

screen shot 2018-06-07 at 9 59 00 am

### Current behavior

screen shot 2018-06-07 at 9 58 40 am

### Possible solution

not sure

1. open IE11
2. go to http://mineral-ui.com/components/start-end/basic

<details><summary>YourFileName.js</summary><p>

```js
// Please paste your code here
```
</p></details>

### Version

0.33

### Environment

prod

### LINKED ISSUE #711: [FlexItem] Cannot easily be a flex container
### Expected behavior

While building a layout, I should be able to apply flex container styles to a FlexItem, so that I can make non-trivial things.

### Current behavior

FlexItem cannot easily be a flex container (inline styles are always a non-ideal option), which prevents making anything but the most trivial of layouts without resorting to a bunch of nested Flex & FlexItems.

### Possible solution

Instead of styling Box to make FlexItem, we could style Flex. Need to consider if it's OK & expected to make every FlexItem also a Flex, though.

1. Render the code below somewhere
2. Expect the inner-most FlexItem to be vertically-centered
3. Be disappointed

<details><summary>App.js</summary><p>

```jsx
<Flex height={100}>
  <FlexItem alignItems="center">
    I'm aligned to the top :(
  </FlexItem>
</Flex>
```
</p></details>

[Edit Mineral UI FlexItem issue repro](https://codesandbox.io/s/5x5v43y85k)

### Context

Came up while observing someone build an app layout with Mineral UI.

### Version

v0.33.0

kylegach:
### Description

- Refactor Flex's gutters
    - To better respect RTL and responsive properties
    - To remove the outer negative margins, which had caused layout issues for consumers and was a prerequisite for...
- Add a `flex` prop to FlexItem to compose Flex instead of Box, which makes it much easier to nest flex containers and flexed items for layout purposes
- Simplify the demo-only styles for Flex, which fixed #710 
- Add some specs for critical/complex functionality
- Update snapshots
    - Pagination snapshots were affected because it uses Flex components
    - Table snapshots were affected because the controlled examples use Flex components

### Motivation and context

Closes #706, closes #710, closes #711

https://flex-improvements--mineral-ui.netlify.com/components/flex

### How to test

2. Navigate to [Flex](https://flex-improvements--mineral-ui.netlify.com/components/flex
) and [FlexItem](https://flex-improvements--mineral-ui.netlify.com/components/flex-item)
3. Confirm everything looks correct, specifically noting the new [flex props example](https://flex-improvements--mineral-ui.netlify.com/components/flex-item/flex-props)

### Types of changes

Breaking: The removal of the negative margins on Flex could have impact on consumer's layout, so I marked it as a breaking change.

### Checklist

Flex

### DIFF
--- a/src/library/Flex/Flex.js
+++ b/src/library/Flex/Flex.js
@@ -1,7 +1,10 @@
 /* @flow */
 import React, { Children, cloneElement } from 'react';
 import { withTheme } from '../themes';
-import { createStyledComponent, getResponsiveStyles, pxToEm } from '../styles';
+import { createStyledComponent, pxToEm } from '../styles';
+import getResponsiveStyles, {
+  getPrevNonNull
+} from '../styles/getResponsiveStyles';
 import Box from '../Box';
 
 type Props = {
@@ -63,60 +66,123 @@ export type Values = boolean | null | number | string;
 const getAlignment = (value: string): string =>
   ['start', 'end'].indexOf(value) !== -1 ? `flex-${value}` : value;
 
-const getGutterSize = (theme: Object, value: number | string): string =>
-  typeof value === 'number'
-    ? pxToEm(value / 2)
-    : `${parseFloat(theme[`space_inline_${value}`] || value) / 2}em`;
-
-const getMargin = (
-  gutter: string,
+const getGutterSize = (
   theme: Object,
   value: number | string
-): string =>
-  value ? `calc(${getMeasurement(theme, value)} - ${gutter})` : `-${gutter}`;
+): number | string =>
+  typeof value === 'number' && value !== 0
+    ? pxToEm(value)
+    : theme[`space_inline_${value}`] || value;
 
-const getMeasurement = (theme: Object, value: number | string): string => {
-  return (
-    theme[`space_inline_${value}`] ||
-    (typeof value === 'number' ? `${value}px` : value)
-  );
-};
+const getIndexedValue = (property, index) =>
+  property && Array.isArray(property) && index !== undefined
+    ? property[index]
+    : property;
 
 const getJustification = (value: string): string =>
   ['around', 'between', 'evenly'].indexOf(value) !== -1
     ? `space-${value}`
     : getAlignment(value);
 
+const getMarginOrGutter = ({
+  gutterWidth,
+  index,
+  margin,
+  marginEnd,
+  marginHorizontal,
+  marginStart,
+  start,
+  theme
+}) => {
+  return (
+    getIndexedValue(start ? marginStart : marginEnd, index) ||
+    getIndexedValue(marginHorizontal, index) ||
+    getIndexedValue(margin, index) ||
+    getGutterSize(theme, gutterWidth)
+  );
+};
+
+const getMarginProps = ({ direction, gutterWidth, theme, ...restProps }) => {
+  if (Array.isArray(direction)) {
+    return direction.reduce(
+      (acc, _, index) => {
+        const value = getPrevNonNull(direction, index);
+
+        if (value === 'row' || value === 'row-reverse') {
+          pushMarginProps({
+            direction: value,
+            index,
+            gutterWidth,
+            props: acc,
+            theme,
+            ...restProps
+          });
+        } else {
+          pushMarginProps({
+            index,
+            gutterWidth: 0,
+            props: acc,
+            theme,
+            ...restProps
+          });
+        }
+        return acc;
+      },
+      { marginStart: [], marginEnd: [] }
+    );
+  } else if (direction === 'row' || direction === 'row-reverse') {
+    const flip = direction === 'row-reverse';
+    const marginProperty = flip ? 'marginStart' : 'marginEnd';
+
+    return {
+      [marginProperty]: getMarginOrGutter({
+        ...restProps,
+        gutterWidth,
+        start: flip,
+        theme
+      })
+    };
+  }
+};
+
+const pushMarginProps = ({
+  direction,
+  index,
+  gutterWidth,
+  props,
+  theme,
+  ...restProps
+}) => {
+  const flip = direction === 'row-reverse';
+  props.marginEnd.push(
+    getMarginOrGutter({
+      ...restProps,
+      gutterWidth: flip ? 0 : gutterWidth,
+      index,
+      theme
+    })
+  );
+  props.marginStart.push(
+    getMarginOrGutter({
+      ...restProps,
+      gutterWidth: flip ? gutterWidth : 0,
+      index,
+      start: true,
+      theme
+    })
+  );
+};
+
 const styles = {
   root: ({
     breakpoints,
     alignItems,
     direction,
-    gutterWidth,
     inline,
-    margin,
-    marginEnd,
-    marginHorizontal,
-    marginLeft: propMarginLeft,
-    marginRight: propMarginRight,
-    marginStart,
     justifyContent,
     theme,
     wrap
   }) => {
-    const rtl = theme.direction === 'rtl';
-    const gutter = getGutterSize(theme, gutterWidth);
-    const marginLeft =
-      propMarginLeft ||
-      (rtl ? marginEnd : marginStart) ||
-      marginHorizontal ||
-      margin;
-    const marginRight =
-      propMarginRight ||
-      (rtl ? marginStart : marginEnd) ||
-      marginHorizontal ||
-      margin;
-
     const mapValueToProperty = (
       property: string,
       value: Values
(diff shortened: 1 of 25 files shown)


kylegach:
There are a number of happo diffs, but I believe they are a result of either the lack of negative margins on Flex (expected) or the newly-simplified demo-only Flex styles (expected and inconsequential to the actual component).

src/website/app/demos/Flex/examples/FlexItem/flexProps.js
```diff
@@ -0,0 +1,36 @@
...
+    display: flex ? 'flex' : 'block',
+    justifyContent: justifyContent || 'flex-start'
+  })
+);
+
+export default {
+  id: 'flex-props',
+  title: 'Flex Props',
+  description: `FlexItem can optionally style the [Flex](/components/flex)
+component with the \`flex\` prop, allowing it to accept any of
```
VictoriaVasys: You may want to call this out more clearly; maybe "The `flex` prop renders the `FlexItem` as a `Flex` container... "

src/website/app/demos/Flex/index.js
```diff
@@ -45,6 +45,10 @@ powerful and flexible enough to be the foundation of your app's layout.`
           In addition to the props above, FlexItem also accepts all props for{' '}
           <a href="/components/box" key={0}>
             Box
+          </a>{' '}
+          and, if <code key={1}>flex</code> is <code key={2}>true</code>,{' '}
```
VictoriaVasys: Since you use "for Box" above, it would be a little more clear with a corresponding "for Flex" here.

src/library/Flex/Flex.js
```diff
@@ -63,60 +66,123 @@ export type Values = boolean | null | number | string;
...
+
+const pushMarginProps = ({
+  direction,
+  index,
+  gutterWidth,
+  props,
+  theme,
+  ...restProps
+}) => {
+  const flip = direction === 'row-reverse';
```
VictoriaVasys: This could be pulled out to decrease duplication; small change though, not critical at all

src/library/Flex/Flex.js
```diff
@@ -63,60 +66,123 @@ export type Values = boolean | null | number | string;
...
+  marginHorizontal,
+  marginStart,
+  start,
+  theme
+}) => {
+  return (
+    getIndexedValue(start ? marginStart : marginEnd, index) ||
+    getIndexedValue(marginHorizontal, index) ||
+    getIndexedValue(margin, index) ||
+    getGutterSize(theme, gutterWidth)
```
VictoriaVasys: I'm a little confused about the logic here; if there is a marginStart or marginEnd (& index), then there is never a marginHorizontal? It might be helpful to discuss in person.

src/website/app/demos/Flex/examples/FlexItem/flexProps.js
kylegach: Updated like so: https://github.com/mineral-ui/mineral-ui/blob/f44b50e937a942bcb33651b8891e64536e34a8ba/src/website/app/demos/Flex/examples/FlexItem/flexProps.js#L20-L23

kylegach: I also updated all of the box props examples to use the same "composes" language: https://github.com/mineral-ui/mineral-ui/blob/f44b50e937a942bcb33651b8891e64536e34a8ba/src/website/app/demos/Flex/examples/FlexItem/boxProps.js#L8-L11

(Only FlexItem's has that second sentence.)

src/website/app/demos/Flex/index.js
kylegach: I tried to make it even more clear, by bringing Flex, the component, before `flex`, the prop: https://github.com/mineral-ui/mineral-ui/blob/f44b50e937a942bcb33651b8891e64536e34a8ba/src/website/app/demos/Flex/index.js#L45-L52

src/library/Flex/Flex.js
kylegach: The `direction` referenced here is a value in an array, whereas the other declaration of `flip` uses the non-array, string value, so the best I could do is make it a function, which seems overkill.

I briefly considered it anyway, as it would be nice to have that `'row-reverse'` detail in one spot, but quickly realized I'd need to refactor much of the file to be consistent in such an approach, as I regular compare against "special strings" like that. So I abandoned that. 😅

src/library/Flex/Flex.js
kylegach: It's an order of precedence thing. If a user provides a marginStart/End, marginHorizontal, or margin prop, the top line (in the code above) wins over the next and over the next until it falls back to applying the gutter.

src/website/app/demos/Flex/components/Flex.js
```diff
@@ -1,41 +1,15 @@
...
-      top: -4,
-      zIndex: -1
-    }
-  };
-};
-
-export default createStyledComponent(_Flex, (props) => ({
-  ...containerStyles(props)
-}));
+export default createStyledComponent(_Flex, (props) => containerStyles(props));
```
brentertz: This could be shortened by removing the extra function wrapper.

```js
export default createStyledComponent(_Flex, containerStyles);
```

src/website/app/demos/Grid/components/Grid.js
```diff
@@ -1,41 +1,6 @@
...
-      top: -4,
-      zIndex: -1
-    }
-  };
-};
-
-export default createStyledComponent(_Grid, (props) => ({
-  ...containerStyles(props)
-}));
+export default createStyledComponent(_Grid, (props) => containerStyles(props));
```
brentertz: Again, this could be shortened by removing the extra function wrapper.

```js
export default createStyledComponent(_Grid, containerStyles);
```

src/library/Flex/Flex.js
```diff
@@ -159,33 +221,31 @@ const Root = createStyledComponent(Box, styles.root, {
...
     let flexItems;
-    flexItems = Children.map(children, (child) => {
-      const {
-        breakpoints: propBreakpoints,
-        margin: propMargin,
-        marginHorizontal,
-        marginLeft,
-        marginRight
-      } = child.props;
+    flexItems = Children.map(children, (child, index) => {
```
brentertz: Can't this just be the following?

```js
const flexItems = Children.map...
```

src/library/Flex/Flex.js
```diff
@@ -159,33 +221,31 @@ const Root = createStyledComponent(Box, styles.root, {
...
+      const { breakpoints: propBreakpoints, ...restChildProps } = child.props;
       let props = {
         breakpoints: propBreakpoints || breakpoints
       };
 
-      if (gutterWidth) {
-        const margin =
-          marginHorizontal || propMargin || getGutterSize(theme, gutterWidth);
+      const flexItemsCount = Children.count(children);
```
brentertz: Can this be moved out of the iteration so it only runs once?

src/library/Flex/__tests__/Flex.spec.js
```diff
@@ -21,4 +47,39 @@ describe('Flex', () => {
 
     expect(flex.exists()).toEqual(true);
   });
+
+  fdescribe('gutter', () => {
```
brentertz: Don't forget to remove the `f`.

src/library/Flex/Flex.js
kylegach: Good call. I had left it inside the `if (child.props)` conditional to avoid the work, but it's better to do it a possible extra one time rather than `n - 1` times.

src/library/Flex/__tests__/Flex.spec.js
kylegach: 🤦‍♂️ . Thanks.

PR #751, opened 2018-07-31
[Pagination]: New component

### LINKED ISSUE #147: [Pagination] New component

VictoriaVasys:
### Description
* Add support for size html attribute with htmlSize prop in TextInput
* Export defaultSortComparator in Table
* Add Pagination component
* Add Pagination website examples
* Add  Table website examples for sortable and selectable paginated tables

### Motivation and context
Closes #147 

https://147-pagination--mineral-ui.netlify.com/

### How to test
2. Navigate to Pagination
3. Confirm all looks and works well
4. Navigate to Table
5. Confirm sortable-paginated and selectable-paginated examples are working appropriately

### Types of changes

Other: See details in Description above

### Checklist
* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support) (#753 will fix IE 11 layout bug)

Pages turning in 3D from an e-reader


src/library/Pagination/PageJumper.js
```diff
@@ -0,0 +1,70 @@
...
+      <Root {...rootProps}>
+        <FormField {...inputProps} />
+      </Root>
+    );
+  }
+
+  handlePageJump = (event: SyntheticInputEvent<>) => {
+    const { onPageChange, totalPages } = this.props;
+    event.persist();
+    const updateCurrentPage = () => {
```
kylegach: Maybe this could be defined as another method rather than created inside this one?

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,156 @@
...
+  totalPages: number,
+  visibleRange: number
+};
+
+const Container = createStyledComponent(FlexItem, ({ theme: baseTheme }) => {
+  const theme = componentTheme(baseTheme);
+  return {
+    '& > button, & > span': {
+      '&:not(:last-child)': {
+        marginRight: theme.Pagination_gutterWidth
```
kylegach: I think this would need the rtl treatment.

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,156 @@
...
+
+const GenericIncrementButton = ({
+  currentPage,
+  direction,
+  handleIncrement,
+  messages,
+  size,
+  totalPages
+}: Props & { direction: string }) => {
+  const incrementForward = direction === 'next' ? true : false;
```
kylegach: Might make sense to do something like `const next = direction === 'next'` and reference that throughout.

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,156 @@
...
+  const disabled =
+    direction === 'next'
+      ? lastPage(currentPage, totalPages)
+      : firstPage(currentPage);
+
+  const iconPosition = direction === 'next' ? 'iconEnd' : 'iconStart';
+
+  const buttonProps = {
+    children: messages.pagination[direction],
+    disabled: disabled,
```
kylegach: Just `disabled,` is needed here.

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,156 @@
...
+    size,
+    onClick: handleIncrement.bind(null, incrementForward)
+  };
+  return <Button {...buttonProps} />;
+};
+
+export default class Pages extends PureComponent<Props> {
+  render() {
+    const { hidePageNumbers, ...restProps } = this.props;
+    const IncrementButton = ({ direction }: { direction: string }) => {
```
kylegach: If you pass `restProps` to both uses of IncrementButton below, you can define this outside of `render()`, which is preferable.

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,245 @@
...
+        label: 'Jump to page',
+        placeholder: 'Page #'
+      },
+      pageSizer: {
+        status: (category, first, last, total) =>
+          `${first}–${last} of ${total} ${category}`,
+        itemText: (pageSize) => `${pageSize} per page`
+      }
+    },
+    pageSizes: [10, 20, 25],
```
kylegach: I wonder if 10, 25, 50 would be better defaults?

src/library/Pagination/PageJumper.js
VictoriaVasys: I did give that a try, but then I have to pass `event` and I couldn't find an event type (or combination of types) that didn't create Flow errors; @brentertz do you have any idea what could work? I tried:

* `SyntheticInputEvent<>`
* `SyntheticInputEvent<> | KeyboardEvent`
* `SyntheticInputEvent<> & KeyboardEvent`
* `SyntheticUIEvent<>`

src/library/Pagination/Pages.js
VictoriaVasys: Yes, thanks for the suggestion! When that is extracted, it actually becomes unnecessary :)

src/website/app/demos/Pagination/examples/table.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        handlePageChange(currentPage) {
+          this.setState({ currentPage });
+        }
+
+        render () {
+          const { currentPage } = this.state;
+          const firstRow = (currentPage - 1) * pageSize;
+          const lastRow = (currentPage - 1) * pageSize + pageSize;
```
VictoriaVasys: @brentertz We thought that indexing currentPage at 1 would improve consumer experience because then it matches what's displayed; our internal code is also simpler. But then it requires the dev to calculate spliced data based on `currentPage - 1`; thoughts on whether we should change it to be 0-indexed?

src/library/TextInput/__tests__/TextInput.spec.js
```diff
@@ -1,12 +1,13 @@
...
-import { shallow } from 'enzyme';
 import TextInput from '../TextInput';
 import examples from '../../../website/app/demos/TextInput/examples';
 import testDemoExamples from '../../../../utils/testDemoExamples';
+import { mountInThemeProvider } from '../../../../utils/enzymeUtils';
 
-function shallowTextInput() {
-  return shallow(<TextInput />);
+function mountTextInput(props = {}) {
+  const [, textInput] = mountInThemeProvider(<TextInput {...props} />);
```
VictoriaVasys: @brentertz For some reason, these tests fail if we don't mount the component. A shallow render is fine for the `TextInput `changes alone, but when we add the Pagination commits, mounting is necessary. Any ideas on why or whether there's a way to change it back?

src/library/Pagination/PageJumper.js
```diff
@@ -0,0 +1,70 @@
...
+import { createStyledComponent } from '../styles';
+import { FlexItem } from '../Flex';
+import { FormField } from '../Form';
+import TextInput from '../TextInput';
+import type { Messages } from './Pagination';
+
+type Props = {
+  'aria-label'?: string,
+  currentPage: number,
+  inputRef: (node: HTMLInputElement) => HTMLInputElement,
```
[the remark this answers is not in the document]
brentertz: Methinks that the TextInput component's innerRef flow type, and that within other components should most likely be defined with the DOM element as the node rather than React$Component.  We would of course need to verify that all of the refs were actually applied this way.  Regardless, try these updates in your `PageJumper`.  They seem to work for me locally.

```js
pageJumperInput: ?HTMLInputElement;
...
inputRef: this.setPageJumperRef,
...
setPageJumperRef = (node: ?HTMLInputElement) => {
  this.pageJumperInput = node;
};
```

src/library/TextInput/__tests__/TextInput.spec.js
brentertz: The shallow tests are very basic smoke tests.  They may not actually render html.  In this case it is likely due to the fact that you need a theme.  Perhaps you could render in a ThemeProvider and then use `.dive()`, but really, I'd just use mount.

...pushed a minor update

src/library/Pagination/PageJumper.js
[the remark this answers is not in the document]
brentertz: Oh, I just pushed the udpate...

brentertz:
Component status page needs updated

src/website/app/demos/Pagination/examples/basic.js
```diff
@@ -0,0 +1,46 @@
+/* @flow */
+import { Component } from 'react';
+import Pagination from '../../../../../library/Pagination';
+
+export default {
+  id: 'basic',
+  title: 'Basic Usage',
+  description: `Pass Pagination a \`currentPage\` to render, a \`pageSize\`
+(number of items or data to be rendered), an \`onPageChange\` handler, and the
```
brentertz: This paragraph is hard to read.  Suggest to remove "items or data", "data or items" and just use "items".

src/website/app/demos/Pagination/examples/visibleRange.js
```diff
@@ -0,0 +1,49 @@
...
+  id: 'visible-range',
+  title: 'Visible Range',
+  description: `The width of the page navigation can be controlled with the
+\`visibleRange\` prop. The value represents the length of the range to be
+displayed between ellipses when the full range of pages is truncated. The value
+should be an odd number, by nature of the design.`,
+  scope: { Component, Pagination },
+  source: `
+    () => {
+      const visibleRange = 3
```
brentertz: Can we just inline this value in the JSX?  The indirection makes the example more complex than it needs to be.

src/library/Pagination/PageJumper.js
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    if (!messages.pageJumper) {
+      throw new Error(
+        'Must provide messages.pageJumper when pageJumper={true}'
+      );
+    }
+
+    const { label, placeholder } = messages.pageJumper;
+    const inputProps = {
```
brentertz: Can we use a number input here that only accepts integers?

src/website/app/demos/Pagination/examples/index.js
```diff
@@ -0,0 +1,26 @@
...
+import visibleRange from './visibleRange';
+import visibleRanges from './visibleRanges';
+
+export default [
+  importSyntax,
+  basic,
+  hidePageNumbers,
+  visibleRange,
+  sizes,
+  pageJumper,
```
brentertz: PageJumper and PageSizer examples should prolly follow one another.

src/website/app/demos/Pagination/examples/kitchenSink.js
```diff
@@ -0,0 +1,80 @@
...
+    columns,
+    data,
+    DemoLayout,
+    Pagination,
+    Table
+  },
+  source: `
+    () => {
+      const pageSizes = [2, 3, 4];
+      const category = 'minerals';
```
brentertz: Again, having these way up here seems to make the example feel more complex.  I'd suggest to simplify by inlining the values in the JSX and just hardcoding a pageSize for the initial state.

src/website/app/demos/Pagination/examples/pageSizer.js
```diff
@@ -0,0 +1,84 @@
...
+import Table from '../../../../../library/Table';
+import DemoLayout from '../../shared/DemoLayout';
+import { columns, data } from '../shared/data';
+
+export default {
+  id: 'page-sizer',
+  title: 'Page Sizer',
+  description: `Allow users to select the page size with the \`pageSizer\` prop.
+A custom collection of optional page sizes can be provided with the
+\`pageSizes\` prop. The category descriptor (e.g. "minerals", below) within the
```
brentertz: I don't think that the `category` prop is going to work for i18n/l10n reasons.  We cannot simply assume that other languages can just append the category value.  We need to give the user full control of the entire string, and provide a sensible default.  I'd suggest adding a function to the messages prop to support this case.

src/website/app/demos/Pagination/examples/visibleRanges.js
```diff
@@ -0,0 +1,76 @@
+/* @flow */
```
brentertz: I am not really sure what this example is demonstrating.  Can it perhaps be turned into a unit test instead?  Alternatively, it needs a description and/or to show the code.

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,247 @@
...
+  /** Called when current page is changed */
+  onPageChange: (currentPage: number) => void,
+  /** Called when [page size](#page-sizer) is changed */
+  onPageSizeChange?: (pageSize: number) => void,
+  /**
+   * Render a [page jumper](#page-jumper)
+   * ([TextInput component](/components/text-input)); enables the user to change
+   * the current page on number entry
+   */
+  pageJumper?: boolean,
```
brentertz: `pageJumper` as a boolean prop feels strange.  I would expect it to take a component or a render function.  I would recommend to rename it to something like `showPageJumper`.  Of course that begs the question of whether `hidePageNumbers` should be renamed to `showPageNumbers` for consistency.

src/library/Pagination/PageJumper.js
```diff
@@ -0,0 +1,70 @@
...
+      </Root>
+    );
+  }
+
+  handlePageJump = (event: SyntheticInputEvent<>) => {
+    const { onPageChange, totalPages } = this.props;
+    event.persist();
+    const updateCurrentPage = () => {
+      const currentPage = parseInt(event.target.value);
+      if (1 <= currentPage && currentPage <= totalPages) {
```
brentertz: Opinion:  I find this line strange to read.  `1 <= currentPage` vs.  `currentPage > 1`

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,151 @@
+/* @flow */
```
brentertz: Some thoughts on focus management...

* If click previous and hit first page, button becomes disabled and focus is lost
* If click next and hit last page, button becomes disabled and focus is lost
* Should we manage focus in these scenarios and set focus somewhere else, like the opposite button?

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,247 @@
...
+        label: 'Jump to page',
+        placeholder: 'Page #'
+      },
+      pageSizer: {
+        status: (category, first, last, total) =>
+          `${first}–${last} of ${total} ${category}`,
+        itemText: (pageSize) => `${pageSize} per page`
+      }
+    },
+    pageSizes: [10, 20, 50],
```
brentertz: Would users ever need to specify an "all" option, in which case, this would need to accept an object of key/value pairs.  I would prolly vote for "not MVP".

src/website/app/demos/Pagination/examples/index.js
VictoriaVasys: Good advice; since it doesn't make sense for the "Table" example to come after "Page Sizer", we moved that example to Table & re-labelled it "Pagination"

src/website/app/demos/Pagination/examples/pageSizer.js
VictoriaVasys: The user does have full control of the string with `messages.pageSizer.status`; we thought that additionally pulling out the `category` string (which is utilized in `status`) would be useful for English speakers who only want to change the category displayed (otherwise they need to redefine all messages).

src/website/app/demos/Pagination/examples/visibleRanges.js
VictoriaVasys: Yes, probably better as a unit test; it was visually helpful while developing (for the edge-case of range of 1), but may be better off as a test from here on.

src/library/Pagination/Pagination.js
VictoriaVasys: Yes, agreed. The title "Hide Page Numbers" seems to contradict our example then; does "Previous & Next" sound ok?

src/library/Table/Table.js
```diff
@@ -169,6 +167,8 @@ class Table extends Component<Props> {
     titleElement: 'h4'
   };
 
+  static defaultSortComparator = defaultSortComparator;
```
brentertz: Why expose this statically and not as a separate import?

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,248 @@
...
+};
+
+export const componentTheme = (baseTheme: Object) => ({
+  PaginationPageJumper_width: '4.65em',
+  Pagination_gutterWidth: baseTheme.space_inline_sm,
+  ...baseTheme
+});
+
+const styles = {
+  pageSizer: ({ theme }) => {
```
brentertz: If there is only one item in styles, I'd prolly just inline these in the definition of PageSizer.

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    return <Root {...rootProps}>{content}</Root>;
+  }
+
+  setPageJumperRef = (node: ?HTMLInputElement) => {
+    this.pageJumperInput = node;
+  };
+
+  totalPages = (pageSize) => Math.ceil(this.props.totalCount / pageSize);
```
brentertz: Can we maybe rename this to `getTotalPages`, `calcTotalPages` or something similar?

src/library/Pagination/Pagination.js
VictoriaVasys: Yes, great point. I think we will save it for future work.

src/library/Pagination/Pages.js
```diff
@@ -0,0 +1,151 @@
...
+  return <Button {...buttonProps} />;
+};
+
+export default class Pages extends PureComponent<Props> {
+  render() {
+    const { showPageNumbers, ...restProps } = this.props;
+
+    return (
+      <Container {...restProps}>
+        <IncrementButton direction={'previous'} {...restProps} />
```
brentertz: `direction="previous"` and `direction="next"` - no need for `{}`

src/library/Pagination/Pagination.js
```diff
@@ -0,0 +1,247 @@
+/* @flow */
```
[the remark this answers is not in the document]
VictoriaVasys: Discussed in person; we added some more guidance in Best Practices.

src/library/Pagination/Pagination.js
VictoriaVasys: Styles should've also had `root` styles; it's updated now.

Issue #742, opened 2018-07-24
[TextInput] Styling issues


G0L0L:
I'm trying to remove border from TextInput by using this:
`const ThemedTextInput = createThemedComponent(TextInput, {
  TextInput_borderWidth: '0px'
});`

And the code works (input element has 0 border), but in fact border remains visiblebecause of element with class `css-<some id here>-Underlay`.

As you can see on the screen it has its own styles, which remains the same.

![image](https://user-images.githubusercontent.com/1982853/43135380-7d751444-8f45-11e8-8691-bbd32c09c733.png)

kylegach:
Thanks for the bug report, @G0L0L!

I just submitted a PR for review that will fix the issue. Should be released in the next couple days.

A couple additional notes:

- The `border: 0` on the `input` itself is always present and cannot be themed. To have the hover/focus states apply around the entire group of `startIcon`, `prefix`, `input`, `suffix`, and `endIcon`, we use that "underlay" to apply the border style (among others). Your code is correct and will work after this bug fix release; its ours that had a bug.
- If you're providing a value in `px`, you can just use a number `0` instead of including the unit in a string, `'0px'`. Either will work, though.

PR #754, opened 2018-08-06
Theme override tests

### LINKED ISSUE #749: Automated testing for component theme overrides
As a mineral developer, I would like automated tests for component theme overrides, so that I can be sure that things are working properly.

brentertz:
### Description

Adds some tests to help ensure that theme overrides work as expected.

Basically, these tests render a normal component, then render a themed version of the same component with a theme variable overridden to null.  It then asserts that the generated styles have changed, thus indicating that the theme override worked.

I tried to test all of the theme variables by default for each component, but it is not always practical due to the different states that some components can be rendered in.  In such cases, a subset or even a single theme variable override is simply being smoke tested.

### Motivation and context

Closes #749

These tests would have caught #742 but they would not have caught #748 

Not really applicable...

https://749-theme-override-tests--mineral-ui.netlify.com/

### How to test

Review the changes and run the test suite

### Types of changes

  - Add tests

### Checklist


src/library/Avatar/__tests__/Avatar.spec.js
```diff
@@ -4,6 +4,7 @@ import { shallow } from 'enzyme';
 import Avatar from '../Avatar';
 import examples from '../../../website/app/demos/Avatar/examples';
 import testDemoExamples from '../../../../utils/testDemoExamples';
+import testThemeOverrides from '../../../../utils/testThemeOverrides';
```
kylegach: It'd be nice, but not a blocker, to import these like `import { testDemoExamples, testThemeOverrides } from '../../../../utils/testing;` (filename at your discretion).

src/library/Select/__tests__/Select.spec.js
```diff
@@ -95,6 +98,25 @@ describe('Select', () => {
...
   });
 
+  describe('theme overrides', () => {
+    testThemeOverrides(
+      <Select data={data} id="test" isOpen>
+        <button>trigger</button>
+      </Select>,
+      getProcessedComponentThemeKeys(componentTheme, {
+        excludeKeys: [
+          'Select_color',
```
kylegach: I didn't see a reason to exclude this one.

utils/testThemeOverrides.js
```diff
@@ -0,0 +1,68 @@
...
+        not applicable in the state in which the component was rendered or it
+        could be symptomatic of a problem with the component implementation.
+      `);
+  }
+}
+
+/**
+ * This function compares the generated styles for the supplied component to a
+ * themed version of the same component wherein each supplied themeKey has been
+ * overridden in a separate expectaion.  Expectations will fail when there is
```
kylegach: expectaion -> expectation

src/library/Select/__tests__/Select.spec.js
kylegach: So, this was a fun little rabbit hole. I checked out this branch, commented out this line, and as expected, got a failing test. Upon applying `showDiffs: true` and inspecting the diff, I found `color: teal` applied to the trigger. Sure enough:

https://github.com/mineral-ui/mineral-ui/blob/master/src/library/Select/SelectTrigger.js#L139

(Which has been there since that file was added, oops: https://github.com/mineral-ui/mineral-ui/pull/581/files#diff-3568d1ea330224d13c62361ce907dc78R128)

That style is on Trigger, which is used as the "control" in FauxControl, which has its own `color` applied (https://github.com/mineral-ui/mineral-ui/blob/master/src/library/FauxControl/FauxControl.js#L140-L159). I'm pretty sure the FauxControl style clobbers the Trigger one, which is confirmed by the resulting teal color not showing up in dev tools.

But the FauxControl style references `theme.FauxControl_color` (which is mapped to `theme.TextInput_color`, which is mapped to `theme.Select_color`), so I'm not sure why updating `theme.Select_color` would have no effect.

Very likely not an issue with this PR, but it deserves further investigation, I think.

brentertz: Good catch, though I could not trace why the override was not working.  I added a FIXME to the test, removed the `color: teal` style from SelectTrigger, and will file a new issue.

PR #759, opened 2018-08-08
[Table] Improve sorting

### LINKED ISSUE #752: [Table] New property request
As a consumer of Table components, I would like to define `rowsLimit` property, which will limit maximum number of rows to render (even if data has more entries).

The scenario:
I have data consists of 100 rows, but I don't want to render them at once. Instead of this I'm going to render it in chunks. Each chunk has 20 rows. So from begging I'm passing to the component data array consisted of 20 rows, by clicking on "Show more" button next 20 rows are added to data and so on.

The problem:
If I have set sorting, it obviously sorts only the data array I passed to component. And when new rows are added, they can appear on any place (since my original data is not sorted).

The solution:
With `rowsLimit` property I no longer need to make data chunks, instead of this I'm passing all my data with a limit. So the component can sort all data and render exact number of sorted rows.

kylegach:
### Description

- Prevent Tables with a controlled `sort` from sorting the data internally (see breaking change, below)
- Update examples & docs necessarily
- Fix a small visual issue with sortable header cells where the dotted border was displaying over the focus outline

### Motivation and context

Closes #752 

https://752-sortable-improvements--mineral-ui.netlify.com/components/table/sortable-paginated

### How to test

2. Navigate to Table > [Sortable — Controlled](https://752-sortable-improvements--mineral-ui.netlify.com/components/table/sortable-controlled) & [Sortable — Paginated](https://752-sortable-improvements--mineral-ui.netlify.com/components/table/sortable-paginated) examples
3. Confirm it works properly

### Types of changes

Breaking: Tables with a controlled `sort` will now need to sort their data and pass that sorted data to the Table

### Checklist

### DIFF
--- a/src/website/app/demos/Table/examples/sortable.js
+++ b/src/website/app/demos/Table/examples/sortable.js
@@ -2,6 +2,12 @@
 import Table from '../../../../../library/Table';
 import sharedData from '../shared/data';
 
+export const sortByLength = (a: Object, b: Object, key: string) => {
+  const lengthA = a[key].length;
+  const lengthB = b[key].length;
+  return lengthA < lengthB ? -1 : lengthA > lengthB ? 1 : 0;
+};
+
 export default {
   id: 'sortable',
   title: 'Sortable Columns',
@@ -28,7 +34,7 @@ const defaultSortComparator: SortComparator = (a, b, key) => {
 };
 \`\`\`
 
-This is available as a named export from the component (\`defaultSortComparator\`):
+This is available as a named export from the component:
 
 \`\`\`
 import Table, { defaultSortComparator } from 'mineral-ui/Table';
@@ -40,9 +46,9 @@ Dairy column sorts by length of the string rather than alphabetically.
   scope: { Table, sharedData },
   source: `
     () => {
-      const sortByLength = (a, b, column) => {
-        const lengthA =a[column].length;
-        const lengthB =b[column].length;
+      const sortByLength = (a, b, key) => {
+        const lengthA = a[key].length;
+        const lengthB = b[key].length;
         return lengthA < lengthB ? -1 : lengthA > lengthB ? 1 : 0;
       };
 
@@ -62,7 +68,8 @@ Dairy column sorts by length of the string rather than alphabetically.
           defaultSort={{ key: 'Fruits' }}
           sortable
           title="Foods of the World"
-          hideTitle />
+          hideTitle
+        />
       );
     }`
 };
--- a/src/website/app/demos/Table/examples/sortableControlled.js
+++ b/src/website/app/demos/Table/examples/sortableControlled.js
@@ -2,23 +2,54 @@
 import { Component } from 'react';
 import Button from '../../../../../library/Button';
 import Flex, { FlexItem } from '../../../../../library/Flex';
-import Table from '../../../../../library/Table';
+import Table, { defaultSortComparator } from '../../../../../library/Table';
 import data from '../shared/data';
+import { sortByLength } from './sortable';
 
 export default {
   id: 'sortable-controlled',
   title: 'Sortable Columns — Controlled',
   description: `Table controls its own sort state by default. It can optionally
 be managed by the application as a controlled component via the control prop,
-\`sort\`.`,
-  scope: { Button, Component, Table, Flex, FlexItem, data },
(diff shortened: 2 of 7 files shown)


src/website/app/demos/Table/examples/sortable.js
```diff
@@ -40,9 +46,9 @@ Dairy column sorts by length of the string rather than alphabetically.
   scope: { Table, sharedData },
   source: `
     () => {
-      const sortByLength = (a, b, column) => {
-        const lengthA =a[column].length;
-        const lengthB =b[column].length;
+      const sortByLength = (a, b, key) => {
+        const lengthA =a[key].length;
```
brentertz: 💄spacing

src/website/app/demos/Table/examples/sortableControlled.js
```diff
@@ -28,65 +49,83 @@ be managed by the application as a controlled component via the control prop,
...
+          this.sortDairyDescending = this.sortDairyDescending.bind(this);
           this.reset = this.reset.bind(this);
         }
 
         handleSort(sort) {
           this.setState({ sort });
         }
 
-        sortFruitsAscending() {
+        // Easily memoized...
```
brentertz: Are you really sure that you don't want to just memoize this method?  I am introducing `memoize-one` in an upcoming PR anyway.  The description does state that it is just one way to do it.  If not, perhaps we can be more clear that it SHOULD be memoized.

kylegach: Alright, you won me over. 😝

src/website/app/demos/Table/examples/sortableControlled.js
```diff
@@ -28,65 +61,72 @@ be managed by the application as a controlled component via the control prop,
           this.state = initialState;
 
           this.handleSort = this.handleSort.bind(this);
-          this.sortFruitsAscending = this.sortFruitsAscending.bind(this);
-          this.sortFruitsDescending = this.sortFruitsDescending.bind(this);
           this.reset = this.reset.bind(this);
+          this.sortDairyAscending = this.sortDairyAscending.bind(this);
+          this.sortDairyDescending = this.sortDairyDescending.bind(this);
+
+          this.sort = memoizeOne(sortFn)
```
VictoriaVasys: This seems to be the only place we use this dependency; is it necessary just for an example?

src/website/app/demos/Table/examples/sortablePaginated.js
```diff
@@ -28,60 +30,69 @@ active sort, current page, and page size. Note that the
...
+            { sort },
             () => { this.handlePageChange(1) }
           );
         }
 
         handlePageChange(currentPage) {
           this.setState({ currentPage })
         }
+        
+        // See 'sortable columns — controlled' example for more implementation details
```
VictoriaVasys: I would capitalize here to match the title of the example (and maybe direct to "above" since we can't add a link)

src/website/app/demos/Table/examples/sortable.js
```diff
@@ -40,9 +46,9 @@ Dairy column sorts by length of the string rather than alphabetically.
   scope: { Table, sharedData },
   source: `
     () => {
-      const sortByLength = (a, b, column) => {
-        const lengthA =a[column].length;
-        const lengthB =b[column].length;
+      const sortByLength = (a, b, key) => {
+        const lengthA = a[key].length;
+        const lengthB = b[key].length;
         return lengthA < lengthB ? -1 : lengthA > lengthB ? 1 : 0;
```
VictoriaVasys: Since this functionality is used in multiple places, would it be better to implement the exported const here & add expose it in a separate comment in this example?

src/website/app/demos/Table/examples/sortableControlled.js
brentertz: Since data is an object, I think you'l need to provide your own equality function as a 2nd parameter to `memoizeOne` and use `deepEqual` with that.

src/website/app/demos/Table/examples/sortableControlled.js
kylegach: On second thought, that's not as trivial of an enhancement as I thought. Saving for future work. I did update the text to "should", though.

PR #761, opened 2018-08-13
[Pagination]: Render in all supported React versions

### LINKED ISSUE #760: [Pagination] Breaks in older react versions
### Expected behavior

Expect that Pagination component works in all supported React versions.

### Current behavior

Get the following...

```
Invariant Violation: Buttons(...): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.
```

### Possible solution

Components must provide a wrapping element when returning an array, in older React versions.

`npm run jest:react`

### Context

Trying to do other mineral work and validate that I have not broken anything and this got in the way.

### Version
v0.41.0

brentertz:
### Description

* Updates the Pagination Pages subcomponent - converts the internal Buttons component to a simple function instead of an actual component.  This way it can still just return an array, without the need for an additional wrapper element.  This means that the DOM structure did not need to change.
* Exclude the Pagination demo example snapshot tests from being tested in older React versions as they use the Table component, which uses the context polyfill, which renders differently in different React versions.

### Motivation and context

Closes #760

Not really applicable...

http://760-pagination--mineral-ui.netlify.com

### How to test

1. `npm run jest:react`

### Types of changes

### Checklist

### DIFF
--- a/src/library/Pagination/Pages.js
+++ b/src/library/Pagination/Pages.js
@@ -52,7 +52,7 @@ const EllipsisButton = createThemedComponent(Button, ({ theme }) => ({
   color_disabled: theme.color_theme
 }));
 
-const Buttons = ({
+const getPageButtons = ({
   currentPage,
   handleClick,
   messages,
@@ -164,11 +164,13 @@ const IncrementButton = ({
     size,
     ...restProps
   };
+
   return <Button {...buttonProps} />;
 };
 
 export default class Pages extends PureComponent<Props> {
   previousButton: ?HTMLButtonElement;
+
   nextButton: ?HTMLButtonElement;
 
   render() {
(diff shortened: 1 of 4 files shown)


src/library/Pagination/__tests__/Pagination.spec.js
```diff
@@ -89,7 +89,9 @@ const findPageSizer = (pagination) => {
 };
 
 describe('Pagination', () => {
-  testDemoExamples(examples);
+  testDemoExamples(examples, {
+    contextPolyfill: true
```
VictoriaVasys: Neat. I didn't know this was an option :)

src/library/Pagination/Pages.js
```diff
@@ -182,7 +184,7 @@ export default class Pages extends PureComponent<Props> {
           innerRef={this.setPreviousButtonRef}
           {...restProps}
         />
-        {showPageNumbers && <Buttons {...restProps} />}
+        {showPageNumbers && getPageButtons(this.props)}
```
VictoriaVasys: Good idea, thanks for the update.

PR #789, opened 2018-10-25
[ButtonGroup]: Fix support for styled/themed Button children

brentertz:
### Description

Fixes support for styled/themed Button children

### Motivation and context

Issue reported by user in chat room and illustrated in the following example.

https://codesandbox.io/s/4zvyzy3694

A "children" example has been added to the ButtonGroup page, but it is dev-only and won't show up on Netlify.

https://fix-button-group--mineral-ui.netlify.com/components/button-group

### How to test

1. Check the Happo diff?
2. Checkout the branch locally, and view ButtonGroup's "children" example.
3. Note that the there are no errors in the console, and that colors are appropriate, particularly variants, when styled/themed and used inside of Dropdown

### Types of changes

### Checklist


src/library/utils/children.js
```diff
@@ -2,6 +2,23 @@
 import { Children } from 'react';
 import { toArray } from './collections';
 
+const hasChildren = (child: React$Element<*>): boolean =>
+  Boolean(child && child.props && child.props.children);
+
+const hasComplexChildren = (child: React$Element<*>): boolean =>
+  hasChildren(child) && typeof child.props.children === 'object';
```
kylegach: Is `typeof child.props.children === 'object'` a safe (and clear) enough check here, given that so many JS entities distill down to objects?

brentertz: Good question.  I believe so.  `hasChildren` handles the `null` check and this essentially checks it is not a `string`, `number`, `boolean`.

PR #793, opened 2018-11-06
[Navigation] Add PrimaryNav, SecondaryNav, & NavItem components

### LINKED ISSUE #787: [Navigation] (horizontal) New component
As a Mineral UI consumer, I want a HorizontalNavigation component, so that users have a standard means of navigating throughout the pages/sections of my app.

#### AC
1. Decide on a name
1. Match [design spec](https://preview.uxpin.com/871cc8a4f408b19ba0af468631cd762258212741#/pages/92296189/simulate/sitemap?mode=i)
1. Consider a11y, keyboard, i18n, rtl

#### Other Considerations
- This can very likely be a styled TabList (should change the name to something more generic)
- Keep the [vertical orientation of navigation](https://preview.uxpin.com/871cc8a4f408b19ba0af468631cd762258212741#/pages/90319558/simulate/sitemap?mode=i) in mind when designing the API

kylegach:
### Description
- Prevent disabled non-`button` elements from being focused
- Add `element` prop to MenuItem and allow tabIndex to be overridden
- Export some types from Menu
- Add PrimaryNav and SecondaryNav (made from Navigation), NavOverflowMenu (internal) and NavItem components
- Add examples & demos for PrimaryNav, SecondaryNav, & NavItem
- Add specs & update snapshots

### Motivation and context

Closes #787 

https://787-navigation--mineral-ui.netlify.com/components/primary-nav

### How to test

1. Navigate to [PrimaryNav](https://787-navigation--mineral-ui.netlify.com/components/primary-nav), [SecondaryNav](https://787-navigation--mineral-ui.netlify.com/components/secondary-nav), & [NavItem](https://787-navigation--mineral-ui.netlify.com/components/nav-item) pages
1. Confirm all looks correct with no typos or errors

### Types of changes

Bug fix: Button's with a non-button element used to be focusable

### Checklist

Dog navigating via scooter

### DIFF
--- a/src/library/Menu/MenuItem.js
+++ b/src/library/Menu/MenuItem.js
@@ -1,11 +1,12 @@
 /* @flow */
 import React, { cloneElement, PureComponent } from 'react';
+import memoizeOne from 'memoize-one';
 import { pxToEm } from '../styles';
 import IconDanger from '../Icon/IconDanger';
 import IconSuccess from '../Icon/IconSuccess';
 import IconWarning from '../Icon/IconWarning';
 import {
-  MenuItemRoot as Root,
+  createMenuItemRootNode,
   MenuItemContent,
   MenuItemInner,
   MenuItemSecondaryText,
@@ -13,7 +14,11 @@ import {
 } from './styled';
 
 import { menuItemPropTypes } from './propTypes';
-import type { MenuItemPropGetter, MenuItemProps } from './types';
+import type {
+  MenuItemDefaultProps,
+  MenuItemPropGetter,
+  MenuItemProps
+} from './types';
 
 const variantIcons = {
   danger: <IconDanger size={pxToEm(24)} />,
@@ -24,7 +29,18 @@ const variantIcons = {
 export default class MenuItem extends PureComponent<MenuItemProps> {
   static displayName = 'MenuItem';
 
+  static defaultProps: MenuItemDefaultProps = {
+    element: 'div'
+  };
+
   static propTypes = menuItemPropTypes;
+
+  // Must be an instance method to avoid affecting other instances memoized keys
+  getRootNode = memoizeOne(
+    createMenuItemRootNode,
+    (nextProps: MenuItemProps, prevProps: MenuItemProps) =>
+      nextProps.element === prevProps.element
+  );
 
   render() {
     const {
@@ -42,6 +58,7 @@ export default class MenuItem extends PureComponent<MenuItemProps> {
       });
     }
 
+    const Root = this.getRootNode(this.props, MenuItem.defaultProps);
     const rootProps = this.getItemProps(this.props);
 
     let startIcon = variant !== undefined && variant && variantIcons[variant];
@@ -73,12 +90,12 @@ export default class MenuItem extends PureComponent<MenuItemProps> {
   }
 
   getItemProps: MenuItemPropGetter = (props = {}) => {
-    const { disabled, onClick, render, ...restProps } = props;
+    const { disabled, onClick, render, tabIndex, ...restProps } = props;
 
     return {
       ...(render ? restProps : {}),
       disabled,
-      tabIndex: disabled ? '-1' : 0,
+      tabIndex: tabIndex !== undefined ? tabIndex : disabled ? '-1' : 0,
       ...(!render ? restProps : {}),
       ...(!disabled ? { onClick, onKeyDown: this.onKeyDown } : {})
     };
--- a/src/library/Menu/styled.js
+++ b/src/library/Menu/styled.js
@@ -2,6 +2,9 @@
 import { createStyledComponent, getNormalizedValue, pxToEm } from '../styles';
 import { menuDividerTheme, menuGroupTheme, menuItemTheme } from './themes';
 
+import type { CreateRootNode } from '../styles/types';
+import type { MenuItemDefaultProps, MenuItemProps } from './types';
+
 export const MenuRoot = createStyledComponent(
   'div',
   {},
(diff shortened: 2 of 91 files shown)


src/library/ButtonGroup/ButtonGroup.js
```diff
@@ -195,7 +195,7 @@ const styles = ({ fullWidth, theme: baseTheme, variant }) => {
 // production, as long as it is set statically on each component
 // https://github.com/facebook/react/issues/4915#issuecomment-335803765
 const isButtonComponent = (element: React$Element<*>) => {
-  return /Button/.test(element.type.displayName);
+  return element.type ? /Button/.test(element.type.displayName) : false;
```
kylegach: I'm not sure why this is only necessary now. Because of the way `isButtonComponent` is called as part of `findDeep` ([line 287]()), sometimes the `element` here is `'[text]'`, which doesn't have a `type` property.

brentertz: I guess that makes sense since technically, elements can be strings.  Anyway, you could prolly rewrite this as...

```js
const isButtonComponent = (element: React$Element<*>) =>
  element.type && /Button/.test(element.type.displayName);
```

src/library/Menu/MenuItem.js
```diff
@@ -73,12 +90,12 @@ export default class MenuItem extends PureComponent<MenuItemProps> {
...
 
   getItemProps: MenuItemPropGetter = (props = {}) => {
-    const { disabled, onClick, render, ...restProps } = props;
+    const { disabled, onClick, render, tabIndex, ...restProps } = props;
 
     return {
       ...(render ? restProps : {}),
       disabled,
-      tabIndex: disabled ? '-1' : 0,
+      tabIndex: tabIndex || (disabled ? '-1' : 0),
```
brentertz: I think this needs a bit more adjustment. It is causing a tabIndex of 0 to be added to Select/Dropdown items and breaking tab navigation.

src/library/Menu/styled.js
```diff
@@ -72,84 +75,94 @@ export const MenuGroupTitle = createStyledComponent(
...
+  const { element = defaultProps.element } = props;
+
+  return createStyledComponent(
+    element,
+    // These styles are based off of Button, with significant changes
+    ({ disabled, isHighlighted, theme: baseTheme, variant }) => {
+      let theme = menuItemTheme(baseTheme);
+
+      if (variant) {
+        // prettier-ignore
```
brentertz: 💄 I think this `prettier-ignore` is also ignoring your indentation here.

src/library/Navigation/NavItem.js
```diff
@@ -0,0 +1,54 @@
...
+    'aria-disabled': disabled,
+    'aria-current': selected ? 'page' : undefined,
+    circular: type === INTERNAL_TYPE.pills,
+    'data-index': index,
+    disabled,
+    iconStart: icon,
+    selected,
+    size:
+      type === INTERNAL_TYPE.pills
+        ? 'small'
```
brentertz: Could prolly use Button's `SIZE` const here.

src/library/Navigation/themes.js
```diff
@@ -0,0 +1,115 @@
...
+  PrimaryNavItem_borderColor_active: 'transparent',
+  PrimaryNavItem_borderColor_focus: baseTheme.color_theme_20,
+  PrimaryNavItem_borderColor_hover: 'transparent',
+  PrimaryNavItem_borderColor_selected: baseTheme.color_theme_80,
+  PrimaryNavItem_color: baseTheme.color_themePrimary,
+  PrimaryNavItem_color_disabled: baseTheme.color_theme_40,
+  PrimaryNavItem_color_hover: baseTheme.color_themePrimary,
+  PrimaryNavItem_color_selected: baseTheme.color_themePrimary,
+
+  PrimaryNavItem_backgroundColor_minimal: 'transparent',
```
brentertz: I realize that you have ordered/grouped these keys intentionally.  It does affect the order that they are displayed in the theme variables table on the website too.  Is this your intent as well, or do you think we should sort that data prior to rendering?

src/library/Navigation/types.js
```diff
@@ -0,0 +1,215 @@
...
+import { ALIGN, INTERNAL_TYPE, PREFIX, TYPE } from './constants';
+
+import type {
+  ComponentTheme,
+  ComponentThemeFn,
+  ThemeValue
+} from '../themes/types';
+import type { MenuItemRenderFn, MenuItemType } from '../Menu/types';
+
+// [1] These purposefully reference $Values instead of $Keys
```
brentertz: I am not sure that `$Values` is doing what you expect here.

For example, try adding the following snippet in this file (anywhere after line 20).

```js
// Expectedly causes flow error because "foo" is not a key in `ALIGN`
const foo: Align = 'foo';
console.log(foo);

// Does not cause flow error because `Prefix` type is actually just `string` and not a union of `PREFIX` values.  I am not sure how to get the equivalent values in Flow;
const bar: Prefix = 'bar';
console.log(bar);
```

src/library/Navigation/types.js
```diff
@@ -0,0 +1,215 @@
...
+type Align = $Keys<typeof ALIGN>;
+type InternalType = $Values<typeof INTERNAL_TYPE>; // [1]
+type ItemElement = React$Element<*> | string;
+type Items = Array<NavigationItem>;
+type Messages = {
+  moreLabel: string,
+  moreText: string
+};
+type Prefix = $Values<typeof PREFIX>; // [1]
+type SharedNavigationProps = {
```
brentertz: `BaseNavigationProps` perhaps?

src/library/Navigation/styled.js
```diff
@@ -0,0 +1,169 @@
...
+            '&, &:focus': {
+              backgroundColor:
+                theme[`${prefix}NavItem_backgroundColor${type}_selected`],
+              color: theme[`${prefix}NavItem_color${type}_selected`]
+            }
+          }
+        : undefined),
+
+      // Truncate
+      '&:active > span > span > span > span > span:focus': {
```
brentertz: Selectors here are definitely a bit fragile.

src/library/Navigation/Navigation.js
```diff
@@ -0,0 +1,173 @@
...
+    };
+
+    return (
+      <NavigationRoot {...rootProps}>
+        {this.renderFromData({ prefix, type })}
+      </NavigationRoot>
+    );
+  }
+
+  renderFromData = (typeAndPrefix: Object) => {
```
brentertz: Is this perhaps better named `renderItems`?

src/library/Navigation/Navigation.js
```diff
@@ -0,0 +1,173 @@
...
+
+    return {
+      disabled,
+      element: element || itemElement,
+      index,
+      maxWidth: maxWidth || maxItemWidth,
+      onClick: !disabled
+        ? composeEventHandlers(onClick, this.handleClick)
+        : // If we render `a` with an `href` for items, we need to preventDefault() to "disable" it
+          (event) => {
```
brentertz: Could extract a `preventDefaultEventFn` or similar so a new function is not created each time.

src/website/app/demos/Navigation/common/navPropDocs.js
```diff
@@ -0,0 +1,81 @@
...
+  align: {
+    description: 'Horizontal alignment of items',
+    type: {
+      name: 'union',
+      value: joinQuoted(Object.values(ALIGN))
+    },
+    defaultValue: `'${component.defaultProps.align}'`
+  },
+  children: {
+    description: `Content of ${componentName}, most likely NavItems
```
brentertz: Is this always `NavItems` now?

brentertz:
Here are some other outstanding issues that I have noticed.  _(Feel free to edit this message)_

* [ ] Tabbing through truncated NavItems requires 2 tab keypresses per item, showing the tooltip on the 2nd keypress, and also gets unexpected focus styles. (It seems Truncate is being focused)

* [ ] Tooltips do not seem to apply consistently.  They are fine on initial page load, but after window resize.  I think Truncate may be missing a debounced EventListener for window resize events.

* [ ] Users may need a way to pass modifiers to the NavOverflowMenu/Dropdown

* [ ] PrimaryNav and SecondaryNav Examples and best-practices still navigate to 404 page when click on item at selectedIndex

* [ ] There is a best practice on the Dropdown page that says not to use Dropdown’s for navigation.  Does that need adjustment/clarification so that we are not sending mixed-messages?

* [ ] Not an issue specific to your branch, but the NavOverflowMenu/dropdown appears overtop the website’s secondary navigation when it is fixed at the top.

src/library/Menu/MenuItem.js
kylegach: Thanks for the catch. Changed it to `tabIndex !== undefined ? tabIndex : disabled ? '-1' : 0`

src/library/Navigation/themes.js
kylegach: Good point. I re-ordered them to be alpha-sorted, like the other component themes.

src/library/Navigation/types.js
kylegach: Hmm... it works if we `type` (instead of `const`) the mapping object:

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgAqAogMqFgC8YA3qmGHAHZ4BcYA5APrN4cA09MBgRx23EXA6oAvgG502fGABqAQxgBXPAGcqYACTqtugDwlyAPgWpgwMADk4YPACdXcV6gDGzHRjAANw1tAEZ2Y209am5eDhs7MGJ3Tx8-AOCTACYIkN19DjgMAAs3eNQgA

Thoughts?

src/library/Navigation/styled.js
```diff
@@ -0,0 +1,169 @@
...
+          }
+        : undefined),
+
+      '& > :not(:first-child)': {
+        marginLeft: theme[`${prefix}Nav_gutter`]
+      }
+    };
+  },
+  {
+    filterProps: ['prefix'],
```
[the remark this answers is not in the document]
kylegach: I had left it off of NavigationRoot on purpose, as it could be either "PrimaryNav" or "SecondaryNav" and I thought just "Navigation" might be more confusing than helpful. Thoughts?

I added it to NavItemButton, though.

I think your hunch re: "label" might be right...

src/library/Navigation/styled.js
kylegach: Yep. If we keep using this in many places, always within a Button, we need to just bake it in.

src/website/app/demos/Navigation/common/navPropDocs.js
kylegach: Yeah, or styled NavItems, but I think we don't call that out for other component docs.

src/library/Menu/MenuItem.js
brentertz: Perhaps we can add a couple quick regression tests to Select and Dropdown? E.g. render open, keydown tab, assert active element

src/library/Navigation/types.js
brentertz: Oh wait, I think I found a better way.  Try this...

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzADEB5EsAXjBICMArAUw3wDooAnBhgLwYAoA3qjBgaAQ3YAuMAHJx7GQBphosbmly1M1AF8AlAG50+AJ4AHBsThxKYACQA1MTACuDXAB5TFuFGJkAPhVUYGAwAHU4dgBrXABCTBwCMAA3AEZpImtbTQUjELCAUXZ2KNwwNTAGAA8LZgYAE0S8QhSAJkzsqhkARxcXapkjIA

Also probably relevant to your curiosity - https://kangax.github.io/compat-table/es5/#test-Object.freeze

src/library/Navigation/styled.js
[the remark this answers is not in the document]
brentertz: Personally, I would still add it as the className otherwise ends up as something like `css-1ez53tx`, and with it at least helps you find the right neighborhood, but I don't feel very strongly about it.

src/library/Navigation/Navigation.js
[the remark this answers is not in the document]
brentertz: Perhaps the flow types here improve on their own when the `$Values` flow type thing is resolved.

PR #798, opened 2018-11-19
[Notification] New Component RFC

brentertz:
### Description

RFC for `Notification` component

### Motivation and context

Document approach and solicit feedback prior to implementation

### DIFF
--- a/docs/rfcs/notification-component.md
+++ b/docs/rfcs/notification-component.md
@@ -1,0 +1,90 @@
+# Notification Component
+
+Notifications are inline messages that are part of the content of a UI, and alert users to a specific screen area that requires their action or attention.  They may appear within a page or dialog.  They are not dismissable, and only disappear once the issue has been resolved.
+
+## Usage
+
+### Basic Usage
+
+```javascript
+import React from 'react';
+import Notification from 'mineral-ui/Notification';
+
+export default () => {
+  return (
+    <Notification
+      title="Weekly Report Available"
+      variant="info"
+      actions={
+        text: 'View Report'
+      }>
+      <Text>
+        Your screen time was down 7% last week, for an average of 2 hours, 53 minutes a day.
+      </Text>
+    </Notification>
+  );
+}
+```
+
+## Exports
+
+* `Notification` (default)
+
+## API
+
+### `Notification` API
+
+| Prop | Type | Required | Default | Description |
+| ---- | ---- | -------- | ------- | ----------- |
+| actions | Array<{}> | false | | Array of Button props |
+| children | React$Node | false | | Body content of the Notification |
+| compact | boolean | false | | Place actions on same line as title. Disallows body content - title only. |
+| title | string \| React$Node | true | | Title of the Notification |
+| variant | 'info' \| 'danger' \| 'success' \| 'warning' | false | 'info' | Intent of the Notification |
+| ...rest | any | false | | All other props will be applied to the root element |
+
+## Dependencies
+
+* `Button` component used for actions
+* Not a direct dependency, but it is expected that any content links are built using mineral-ui `Link` component
+
+## Accessibility
+
+* No custom focus management necessary, though any action buttons and/or links contained within must be focusable
+* `role="alert"` applied to root, which implies `aria-live="assertive"`
+* Unsure if need to specify `aria-labelledby` or `aria-describedby` referencing internal title and body elements, or if they will be picked up automatically.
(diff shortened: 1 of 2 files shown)


docs/rfcs/notification-component.md
```diff
@@ -0,0 +1,90 @@
...
+| ...rest | any | false | | All other props will be applied to the root element |
+
+## Dependencies
+
+* `Button` component used for actions
+* Not a direct dependency, but it is expected that any content links are built using mineral-ui `Link` component
+
+## Accessibility
+
+* No custom focus management necessary, though any action buttons and/or links contained within must be focusable
```
kylegach: I don't think this is true for notifications with actions:

"Typically, a message asking a user to do something would form the content of a dialog window (or inline disclosure), and be accompanied by a choice of action buttons. Because the keyboard operator will need to access those buttons, focus must be moved into the dialog.

For the purpose of this article, what I mean by "notification" is a message that just lets you know what's going on. This may be so you can choose to take action later, or it may be to assure you of an event having taken place already.

In screen reader and keyboard accessibility terms, it's important that focus is not moved to such messages. If there is nothing to be done with the tool, you don't put the tool in the person's hand. Despite this, moving focus has endured as a 'best practice'. Why? Because focusing an element has, traditionally, been the most reliable way to get that element and its contents announced in screen readers."

— https://inclusive-components.design/notifications/

kylegach: Maybe notifications with actions, because they'd require a focus strategy, would be a separate component? Maybe we render the correct component internally if `actions` is provided and only expose Notification publicly to simplify the API?

docs/rfcs/notification-component.md
```diff
@@ -0,0 +1,90 @@
...
+| children | React$Node | false | | Body content of the Notification |
+| compact | boolean | false | | Place actions on same line as title. Disallows body content - title only. |
+| title | string \| React$Node | true | | Title of the Notification |
+| variant | 'info' \| 'danger' \| 'success' \| 'warning' | false | 'info' | Intent of the Notification |
+| ...rest | any | false | | All other props will be applied to the root element |
+
+## Dependencies
+
+* `Button` component used for actions
+* Not a direct dependency, but it is expected that any content links are built using mineral-ui `Link` component
```
kylegach: I think, because of the need to _not_ focus notifications (see comment below), the body content cannot contain interactive elements.

Unless, of course, we provide some kind of focus management.

docs/rfcs/notification-component.md
```diff
@@ -0,0 +1,90 @@
...
+
+Notifications are inline messages that are part of the content of a UI, and alert users to a specific screen area that requires their action or attention.  They may appear within a page or dialog.  They are not dismissable, and only disappear once the issue has been resolved.
+
+## Usage
+
+### Basic Usage
+
+```javascript
+import React from 'react';
+import Notification from 'mineral-ui/Notification';
```
kylegach: `import Text from 'mineral-ui/Text';`

docs/rfcs/notification-component.md
brentertz: FYI, This component is intended as an "inline" notice, rather than an alert dialog.

brentertz:
There are several designs for related messaging components (Toast, Inline Message, and Notification) and the differences are not entirely clear.  I think identifying the proper nomenclature would perhaps be the best starting point for us going forward.  e.g. Perhaps this is an Alert or Status component.

FWIW, I hadn’t actually seen the Notification component specs as they were kind of hidden and thus renamed InlineMessage as that is just unwieldy.

Let’s get some clarification from design before proceeding.

docs/rfcs/notification-component.md
brentertz: Gonna put this conversation on hold until we ensure that everyone is using the proper nomenclature for the different messaging components.

docs/rfcs/notification-component.md
brentertz: Gonna put this conversation on hold until we ensure that everyone is using the proper nomenclature for the different messaging components.

brentertz:
Notes here regarding focus management of `alert` contrasted with `alert-dialog`. 
* https://www.w3.org/TR/wai-aria-1.1/#alert
* https://www.w3.org/TR/wai-aria-1.1/#alertdialog

brentertz:
Closing for now.  Will do more research and open a new PR.

PR #802, opened 2018-11-27
[Header] New component RFC

kylegach:
### Description
RFC for Header component

### Motivation and context
Document approach and solicit feedback prior to implementation

### DIFF
--- a/docs/rfcs/header.md
+++ b/docs/rfcs/header.md
@@ -1,0 +1,197 @@
+# Header Component
+
+Header provides a place for an app logo & title, functionality like primary
+navigation or a workspace switcher, as well as global-level actions like search,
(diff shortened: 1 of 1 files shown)


docs/rfcs/header.md
```diff
@@ -0,0 +1,192 @@
...
+- Select
+
+## Accessibility
+
+Identify a11y strategy, e.g. aria, role, keyboard controls, etc.
+
+- Use `header` element at root
+- Skip link (could be a separate component that is optionally included here?)
+- "auto focus" nature of Search input after clicking/tapping Search button
+- Most other a11y provided by subcomponents (PrimaryNav, Dropdown, etc...)
```
brentertz: Consider whether aria `role="search"` might apply.  https://www.w3.org/TR/wai-aria-1.1/#search

Issue #804, opened 2018-11-30
IE11 does not support Array.find


brentertz:
### Expected behavior

Mineral components render properly in IE11

### Current behavior

In IE11, in a fresh CRA app, with the [IE11 polyfill](https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/ie11.js) included, several mineral components will cause console errors and return a WSOD due to usage of `Array.find` in several places (`ButtonGroup`, `PageSizer`, and `utils/children`), which is sadly [not natively supported by IE11](https://caniuse.com/#search=array.find).

### Possible solution

The obvious interim solution is for users to polyfill their application with support for `Array.find` if IE11 support is needed.  This could perhaps be done using [babel-preset-env](https://babeljs.io/docs/en/babel-preset-env), or  [babel-polyfill](https://babeljs.io/docs/en/babel-polyfill), or by importing the module directly from [core-js](https://github.com/zloirock/core-js#ecmascript-array).

Longer term, mineral just needs to take care of #625 

### Steps to reproduce (for bugs)

* Render aforementioned components in a new CRA, include the [IE 11 polyfill](https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/ie11.js), which does not include `Array.find`, as instructed by [react-app-polyfill](https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill).  
* Stare in awe at the nothingness

### Version

0.55.0

PR #805, opened 2018-11-30
Update dependencies

brentertz:
### Description

Update the majority of project dependencies. Some dependencies were skipped for now as they will be separate efforts.  Check out individual commits and https://david-dm.org/mineral-ui/mineral-ui/update-deps for details.

### Motivation and context

Maintenance and updates for longer term strategy of dropping support for older React versions and updating to the latest Emotion

https://update-deps--mineral-ui.netlify.com/

### How to test

1. Should be no changes

### Types of changes

Maintenance

### Checklist

* [x] Renders and functions properly in [all supported browsers](https://github.com/mineral-ui/mineral-ui#browser-support) - Yes, with the exception of the recently discovered #804 


brentertz:
Hello @trotzig 👋 

I can't seem to make Happo happy on this PR and am wondering if you might have any insight.

Mineral was on `happo.io@1.5.1` and I have updated it to `3.3.0` as seemed to be necessary in order for things to work with babel 7.  Everything runs, but the screenshots are entirely expected - almost like there is no CSS being loaded.

I went through the [happo release notes](https://github.com/happo/happo.io/releases) and the only necessary change that I could spot was `getRootElement` -> `rootElementSelector`, which I did.  I also opened a [small PR on happo.io](https://github.com/happo/happo.io/pull/42).  Am I missing something obvious?

trotzig:
@brentertz I'm looking into it right away.

brentertz:
> @brentertz I'm looking into it right away.

Thanks @trotzig , but please don't rush.  This can easily wait until next week.

trotzig:
My suspicion is that this happo.io commit is causing issues: https://github.com/happo/happo.io/commit/d0bf8dd

Before that change, components were never unmounted. Some css-in-js libraries will ignore unmounting, some will not. I suspect Emotion is removing styles from the DOM on umount. Happo grabs css after all component examples are processed, so if styles are removed from the DOM it will never see them .

To work around this, you can try doing what I did in a different project: 

1. add a `setupScript` to `.happo.js`: 
```js
{ 
    setupScript: path.resolve(__dirname, 'setupHappo.js'),
}
```

2. In that `setupHappo.js` file, disable unmounting: 

```js
// Make cleanup a no-op, to avoid ummounting css styles
window.happoCleanup = () => null;
```

brentertz:
@trotzig That really sounded like it could have been it, but it seems to have rendered in the same fashion.

trotzig:
Okay, my second theory is related to how Happo extract styles from the document. IIRC, we grab css from all `<style>` tags by just using `el.innerHTML`. Some css-in-js libs will not put the styles in the DOM directly, instead they'll need to be accessed using `el.sheet.cssRules`. I made a recent change in a Happo plugin, and I think we might have to do something similar in the happo.io client: https://github.com/happo/happo-plugin-scrape/commit/e292139593776d8bb2070b7d73f7064195c86b66

brentertz:
@trotzig That's a good insight. I am currently experimenting with the [speedy flag in Emotion](https://github.com/emotion-js/emotion/blob/master/docs/ssr.md), typically used for SSR, to force the style tags to be rendered to the DOM.  🤞

brentertz:
@trotzig Yay, that seems to have done the trick. Thanks so much for the help! 🎉

PR #807, opened 2018-12-19
Fix for incompatible version

ninjames101:
There have been upgrades to emotion, specifically emotion has moved to @emotion/core and react-emotion has moved to @emotion/styled. Until the new version of emotion is supported here is a fix that works. Thanks for time 🤗.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,13 @@ Install the [Mineral UI package](https://www.npmjs.com/package/mineral-ui) and
 its dependencies.
 
 ```bash
-npm install --save mineral-ui emotion emotion-theming react-emotion react react-dom
+npm install --save mineral-ui@0.55.0 emotion@9.2.1 emotion-theming@9.2.1 react-emotion@9.2.1 react react-dom
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -26,13 +26,13 @@ Install the [Mineral UI package](https://www.npmjs.com/package/mineral-ui) and
 its dependencies.
 
 ```bash
-npm install --save mineral-ui emotion emotion-theming react-emotion react react-dom
+npm install --save mineral-ui@0.55.0 emotion@9.2.1 emotion-theming@9.2.1 react-emotion@9.2.1 react react-dom
```
kylegach: We want to encourage use of the latest version of Mineral, so we shouldn't pin to `@0.55.0`.

kylegach: `9.2.9` is the latest, pre-10 version that is shared across those three dependencies.

chadwithuhc:
@kylegach you can squash and merge in a PR if you choose the down arrow on the merge button. It's a nice new option 🥇

PR #811, opened 2018-12-28
chore(dependencies): Update to emotion v10.0.5

machiaveli88:
### Description
Update to emotion v10.0.5 and replace react-emotion with @emotion/core @emotion/styled.

### Motivation and context
Update emotion to newest version to use css prop with theme like
`<div css={theme => ({ color: theme.colors.primary })}> some other text </div>`
(from [https://emotion.sh/docs/theming](url))

### Checklist
* [ ] Automated tests written and passing


machiaveli88:
Hi Brent, thanks a lot! I'd tried to migrate to emotion v10 but there are two issues left: First of all the theme can not be overwrite (see test textWithThemeOverrides.js), second is the missing styles of dialog (run app & see http://localhost:8080/components/dialog#trigger). Snapshots of Dialog have to be update, too!
There are also many warnings like `'innerRef' is deprecated and will be removed in a future major version of Emotion, please use the 'ref' prop instead in the usage of 'Truncate'` and `The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to "first-of-type"`, but I'm not able to fix them.

For migrating, this will help: [https://emotion.sh/docs/migrating-to-emotion-10](https://emotion.sh/docs/migrating-to-emotion-10)

Let me know if you need any help.

machiaveli88:
Okay, to fix the second issue, you have to wrap Portal with ThemeProvider!

trotzig:
@brentertz @machiaveli88 👋from happo.io

I came by this PR after seeing it on happo.io. I'm trying to track down why Happo says "347 added examples". I was not expecting that. Part of the reason is that an empty report has been created for the baseline commit (e51dd983e9c31588816fcc7444ac6559aea01cfb). My current theory is that a previous PR (https://github.com/mineral-ui/mineral-ui/pull/810) caused it. That PR adds a yarn.lock file (by mistake?), which causes a check in the `happo-ci` script to do the wrong thing, in turn causing an empty report to be uploaded for e51dd983e9. 

I'm going to manually force a proper report for e51dd983e9 which I believe will fix the issue going forward, so **there's no need for action on your side**.

machiaveli88:
hey @trotzig! Yay, I used first yarn, then I realized my rookie mistake!
thanks for your help!

@brentertz, do you know when you're gonna get to look at emotion 10?

brentertz:
Hey @machiaveli88 We are actively working on the Emotion 10 update.  It is a pretty big task though as we really want to capitalize on some of the new capabilities. Follow the [emotion-10 branch](https://github.com/mineral-ui/mineral-ui/compare/emotion-10) if you are interested.

brentertz:
Hi Philipp, thanks again for your efforts on this PR.  I am however going to close it in favor of https://github.com/mineral-ui/mineral-ui/pull/816 which includes the Emotion 10 update as well as a number of other changes.

PR #812, opened 2019-01-08
[Dialog] [Flex] Fix layout issues

brentertz:
### Description

* Add `minWidth` prop to FlexItem
  * Add demo example
* Fix additional layout issues
  * Ensure DialogBody & FlexItem children do not extend beyond element boundary
* Update Dialog Composition example to illustrate layout scenario

### Motivation and context

Fix issues reported by users in chat.

https://layout-bugs--mineral-ui.netlify.com/components/flex-item/min-width

### How to test

1. Navigate to FlexItem > [Min Width example](https://layout-bugs--mineral-ui.netlify.com/components/flex-item/min-width)
1. Confirm that the Select is truncated properly
1. Navigate to Dialog > Composition example (only viewable locally)
1. Confirm that the Select is truncated properly

### Types of changes

- Feature

### Checklist

### DIFF
--- a/src/library/Box/propTypes.js
+++ b/src/library/Box/propTypes.js
@@ -1,7 +1,7 @@
 /* @flow */
 import { arrayOf, bool, number, oneOfType, string } from 'prop-types';
 
-const spacingPropType = oneOfType([
+export const spacingPropType = oneOfType([
   number,
   string,
   arrayOf(oneOfType([number, string]))
--- a/src/library/Box/types.js
+++ b/src/library/Box/types.js
@@ -42,7 +42,7 @@ export type SpacingStyles = {
 
 export type SpacingValue = StyleValue | SpacingSize;
 
-type HeightOrWidthProp = number | string | Array<number | string | null>;
+export type HeightOrWidthProp = number | string | Array<number | string | null>;
 
 type SpacingProp =
   | number
--- a/src/library/Dialog/__tests__/__snapshots__/Dialog.spec.js.snap
+++ b/src/library/Dialog/__tests__/__snapshots__/Dialog.spec.js.snap
@@ -285,6 +285,7 @@ exports[`Dialog demo examples Snapshots: actions 1`] = `
   -webkit-flex: 1 1 auto;
   -ms-flex: 1 1 auto;
   flex: 1 1 auto;
+  width: 100%;
 }
 
 .emotion-25 > div {
@@ -308,6 +309,7 @@ exports[`Dialog demo examples Snapshots: actions 1`] = `
   -webkit-flex: 1 1 auto;
   -ms-flex: 1 1 auto;
   flex: 1 1 auto;
+  width: 100%;
 }
 
 .emotion-24 > div {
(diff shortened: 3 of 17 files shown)


brentertz:
@trotzig We are seeing some unexpected diffs in this PR.  The only component that we expect to have a diff here is FlexItem.  

Here is [a report](https://happo.io/a/14/p/14/report/6fa5d14eee471de32b2020340c8bd2518f27b50e) generated by [a PR](https://github.com/mineral-ui/mineral-ui/pull/807), corresponding to the latest commit on master, that made only doc/website changes (so no diffs expected). Notice that the snapshots contained in that report do not match those being compared against in this PR's report. It would seem that the same baseline is generating different resulting snapshots.  Could this perhaps be due to changes on your end that occurred between that PR and this PR?

trotzig:
@brentertz I think I know what happened here, and I think it's safe to accept that Happo report and move on. Last week I noticed that a PR had caused an empty report to be uploaded by mistake. It was a rare thing caused by a PR accidentally checking in a `yarn.lock` file (which you don't use, but it tripped up Happo). To fix this, I manually created a report for e51dd983e9c31588816fcc7444ac6559aea01cfb (which is what this PR is using as a baseline). That report ended up being slightly different than others which I think is because I hadn't run all required build steps or something. 

There's a little more info here: 
https://github.com/mineral-ui/mineral-ui/pull/811#issuecomment-451146273

This is unfortunate but I think it's rare enough that we don't have to worry too much. I'll try to make this (yarn/npm auto detection) a little more robust next time I'm working on those CI scripts.

kylegach:
@trotzig — Thanks for the explanation. That mostly makes sense, but this part still confuses us:

> Last week I noticed that a PR had caused an empty report to be uploaded by mistake.

That PR (#810) was never merged, so how did it affect our baseline? Is that related to how the unexpected yarn.lock file tripped up Happo?

trotzig:
This will get somewhat deep into technical details, sorry about that. 

The happo CI script will roughly do this for a PR: 
- check if there is a report for the baseline commit (usually the latest commit on master)
- if not, `git checkout` that commit and create a report
- then, `git checkout` the tip of the PR and create a report

In addition to that first step, there's some code in place to handle the "first Happo run" case, meaning there is no baseline report, and happo isn't installed (in package.json) for the baseline commit. In that case, we create an empty report for the baseline which in turn leads to a happo status for the PR saying something like `163 snapshots added`. 

The added `yarn.lock` tricked happo into using `yarn` to generate the baseline commit, but it wasn't working, which in turn led to happo thinking this was the first run. The relevant code is around here: 
https://github.com/happo/happo.io/blob/95c5847c79fedea75701ba35f9c6f1cb9f7b0055/bin/happo-ci#L55

I've made a note about making this more robust. One thing I could do is to do the yarn/npm feature detection every time we make a command (as opposed to once in the start of the script).

trotzig:
I did make some changes to CI scripts recently that would have prevented this from happening. In the past, the `happo-ci` scripts (I believe you use `happo-ci-travis`) used to only run on PRs. With the latest version ([3.6.0](https://github.com/happo/happo.io/releases/tag/v3.6.0)), happo will run on single commits as well. This means most of the time when someone makes a PR, the baseline report will already be there.

kylegach:
@trotzig — Thank you for taking the time to write that up. Makes perfect sense now.

